PR #4, opened 2019-10-20
 Implement Prometheus Metrics

viivek46:
## What changes are proposed in this PR? 
Resolves  [DCOS-59498](https://jira.mesosphere.com/browse/DCOS-59498) 

This PR Implement Prometheus Metrics for kudo Cassandra operator. we have used [criteo cassandra_exporter](https://github.com/criteo/cassandra_exporter) to export Cassandra metrics.

To implement this below decisions are made : 
1. created different container in same pod to run metrics, and added related parameters.
2. added config file cassandra-exporter-config-yml and downloaded jar for criteo cassandra exporter.
3. Added label kudo.dev/servicemonitor: "true" and created  service-monitor.yaml file to create Service Monitor object for service discovery.

## How were these changes tested?
Manually tested and Below are steps to check 
1. Spin up Konvoy cluster and ensure cluster config is applied for kubectl.
2. Install KUDO Cassandra operator
3. Run command to create service-monito: kubectl create -f resources/service-monitor.yaml
4. Wait 10-20 sec, and go to Prometheus dashboard. Type there Cassandra you will see "cassandra_stats".


viivek46:
Hi @mpereira / @samvantran, 
Few things that I need suggestion on : 
1. Metrics CPU/memory and METRICS PORT - what I should set default values.
2. I copied  cassandra-exporter-config-yml.yaml from criteo, should we go ahead with this or any things which we should remove from blacklist, also i am not sure how we can provide the customer option to enter blacklist values. 

currently i am using curl to get cassandra_exporter jar from git, but once all things are ok i will upload this jar to http://downloads.mesosphere.com.

operator/operator.yaml
```diff
@@ -16,6 +16,7 @@ tasks:
       - cassandra-env-sh.yaml
       - jvm-options.yaml
       - stateful-set.yaml
+      - cassandra-exporter-config-yml.yaml
```
samvantran: can we remove `-yml`? Seems redundant to say yaml yaml

operator/params.yaml
```diff
@@ -22,6 +22,22 @@ NODE_MEM_LIMIT_MIB:
   description: "Memory limit (in MiB) for the Cassandra node pods."
   default: 4096
 
+METRICS_CPU_MC:
+  description: "CPU request (in millicores) for the Cassandra node pods."
```
samvantran: Is this for the node pods or metrics pods?

operator/templates/cassandra-exporter-config-yml.yaml
```diff
@@ -0,0 +1,63 @@
...
+       # > java -Dorg.slf4j.simpleLogger.defaultLogLevel=trace -jar cassandra_exporter.jar config.yml --oneshot
+       #
+       # To get intuition of what is done by cassandra when something is called you can look in cassandra
+       # https://github.com/apache/cassandra/tree/trunk/src/java/org/apache/cassandra/metrics
+       # Please avoid to scrape frequently those calls that are iterating over all sstables
+
+       # Unaccessible metrics (not enough privilege)
+       - java:lang:memorypool:.*usagethreshold.*
+
+       # Leaf attributes not interesting for us but that are presents in many path
```
samvantran: Was this copy paste from Criteo's config? Some of these attributes _do_ seem interesting... but maybe it exposes too many metrics.

operator/templates/stateful-set.yaml
```diff
@@ -46,7 +46,7 @@ spec:
           imagePullPolicy: Always
           # Dockerfile based on:
           # https://github.com/docker-library/cassandra/blob/master/3.11/Dockerfile
-          image: mpereira/cassandra:3.11.4
+          image: viivek46/cassandra:3.11.4
```
samvantran: hm, can we add tooling to use non-personal images? and can we add additional versioning so that we don't overwrite each other's images?

operator/templates/stateful-set.yaml
```diff
@@ -133,6 +135,63 @@ spec:
             - name: jvm-options
               mountPath: /etc/cassandra/jvm.options
               subPath: jvm.options
+        {{ if eq .Params.METRICS_ENABLED "true" }}
+        - name: metrics
+          imagePullPolicy: Always
+          image: viivek46/cassandra:3.11.4
```
samvantran: Is it possible to parameterize the image so that we don't hardcode it in multiple places?

resources/service-monitor.yaml
```diff
@@ -0,0 +1,17 @@
...
+    app: prometheus-operator
+    release: prometheus-kubeaddons
+  name: cassandra-cluster-monitor
+spec:
+  endpoints:
+  - interval: 30s
+    port: metrics
+  namespaceSelector:
+    matchNames:
+    - default
```
samvantran: Should we add a param like `{{ .Params.Prometheus.Namespace }}` ? This currently assumes prom is installed under `default` (like it is with Konvoy) but that might not be the case with users on different k8s distros

operator/templates/cassandra-exporter-config-yml.yaml
```diff
@@ -0,0 +1,63 @@
...
+       - .*:mean
+       - .*:min
+
+       # Path present in many metrics but uninterresting
+       - .*:viewlockacquiretime:.*
+       - .*:viewreadtime:.*
+       - .*:cas[a-z]+latency:.*
+       - .*:colupdatetimedeltahistogram:.*
+
+       # Mostly for RPC, do not scrap them
```
samvantran:
```suggestion
       # RPC metrics that do not need to be scraped
```

operator/templates/cassandra-exporter-config-yml.yaml
```diff
@@ -0,0 +1,63 @@
...
+       - .*:fifteenminuterate
+       - .*:fiveminuterate
+       - .*:durationunit
+       - .*:rateunit
+       - .*:stddev
+       - .*:meanrate
+       - .*:mean
+       - .*:min
+
+       # Path present in many metrics but uninterresting
```
samvantran:
```suggestion
       # Non-interesting metrics
```

operator/templates/cassandra-exporter-config-yml.yaml
```diff
@@ -0,0 +1,63 @@
...
+       - org:apache:cassandra:db:.*
+
+       # columnfamily is an alias for Table metrics
+       # https://github.com/apache/cassandra/blob/8b3a60b9a7dbefeecc06bace617279612ec7092d/src/java/org/apache/cassandra/metrics/TableMetrics.java#L162
+       - org:apache:cassandra:metrics:columnfamily:.*
+
+       # Should we export metrics for system keyspaces/tables ?
+       - org:apache:cassandra:metrics:[^:]+:system[^:]*:.*
+
+       # Don't scrap us
```
samvantran:
```suggestion
       # Don't scrape metrics from Criteo
```

operator/templates/cassandra-exporter-config-yml.yaml
```diff
@@ -0,0 +1,63 @@
...
+       - com:criteo:nosql:cassandra:exporter:.*
+
+    maxScrapFrequencyInSec:
```
samvantran: I'm surprised this file had so many typos of the same word...

operator/params.yaml
```diff
@@ -74,6 +90,13 @@ SEED_PROVIDER_CLASS:
   description: "The class within Cassandra that handles the seed logic."
   default: "org.apache.cassandra.locator.SimpleSeedProvider"
 
+METRICS_ENABLED:
+  default: "true"
+
+METRICS_PORT:
+  description: "metrics Port"
```
samvantran:
```suggestion
  description: "Metrics Port"
```

operator/operator.yaml
viivek46: I just followed what been used already like "cassandra-yaml.yaml"
so first part is actual name followed by .yaml

operator/params.yaml
viivek46: updated description for metrics related param.

operator/templates/cassandra-exporter-config-yml.yaml
viivek46: not actually recommended but example :   https://github.com/criteo/cassandra_exporter#config-file-example

mpereira: @samvantran keep in mind that this is the blacklist.

operator/templates/cassandra-exporter-config-yml.yaml
mpereira: https://github.com/criteo/cassandra_exporter/pull/60

operator/templates/stateful-set.yaml
mpereira: We don't need to use the Cassandra image, right?

Here's what I propose in terms of KUDO Cassandra depending on the cassandra_exporter repository:
1. Add cassandra_exporter it as a git submodule under `vendor/cassandra_exporter`
2. Create a symlink from `images/Dockerfile.prometheus-exporter` to `vendor/cassandra_exporter/docker/Dockerfile`
3. Build/push `mesosphere/cassandra_exporter` Docker image based on `images/Dockerfile.prometheus-exporter` (can be personal for now until I finish implementing Docker image build support) and use that for the prometheus exporter container.
4. Revert the changes in the Cassandra node Dockerfile and pod.

operator/templates/stateful-set.yaml
```diff
@@ -133,6 +135,63 @@ spec:
             - name: jvm-options
               mountPath: /etc/cassandra/jvm.options
               subPath: jvm.options
+        {{ if eq .Params.METRICS_ENABLED "true" }}
+        - name: metrics
```
mpereira: In a similar vein to the previous naming comment, let's s/metrics/prometheus-exporter/` everywhere.

operator/templates/stateful-set.yaml
```diff
@@ -133,6 +135,63 @@ spec:
...
+          image: viivek46/cassandra:3.11.4
+          securityContext:
+            capabilities:
+              add:
+                # TODO(mpereira): add an operator parameter for "memory lock" so
+                # that this capability and a respective Docker image with
+                # `setcap`ed binaries are conditionally used.
+                - IPC_LOCK
+                # FIXME(mpereira): might not be necessary given DCOS-59219.
+                - SYS_RESOURCE
```
mpereira: Let's remove this entire `securityContext` block for this container, it doesn't need it.

operator/templates/stateful-set.yaml
```diff
@@ -133,6 +135,63 @@ spec:
...
+          command:
+            - bash
+            - -c
+          args:
+            - cd /etc/cassandra;
+              curl -L -O https://github.com/criteo/cassandra_exporter/releases/download/2.2.1/cassandra_exporter-2.2.1-all.jar;
+              java -jar cassandra_exporter-2.2.1-all.jar cassandra-exporter-config.yml;
```
mpereira: Using a Docker image based on the one available in the cassandra_exporter repository (as suggested in a previous comment) will allow the container `cmd` and `args` to be much simpler and not require downloading the JAR at runtime.

resources/service-monitor.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app: prometheus-operator
+    release: prometheus-kubeaddons
+  name: cassandra-cluster-monitor
```
mpereira:
```suggestion
  name: cassandra-monitor
```

resources/service-monitor.yaml
```diff
@@ -0,0 +1,17 @@
...
+spec:
+  endpoints:
+  - interval: 30s
```
mpereira: This YAML file is not properly indented, please check out: https://github.com/mesosphere/kudo-cassandra-operator/blob/05b09e8127c02e179435975c0b127e432fde134e/README.md#running-static-code-analyzers

resources/service-monitor.yaml
mpereira: Dang, that's a good point Sam. It appears that even Kafka is [setting it statically to `"default"`](https://github.com/kudobuilder/operators/blob/53a70a657b23814a62a04495ffe72fe7869d7666/repository/kafka/docs/latest/resources/service-monitor.yaml#L14).

Let's add `SERVICE_MONITOR_PROMETHEUS_NAMESPACE` in `params.yaml`.

mpereira:
Hey @viivek46, thanks for your initial work! Changes mostly look good, I left a few comments regarding some changes.

Also, we'll probably want an integration test to make sure this actually works. Maybe we can work on it together? Since I'm still wrapping up #3.

operator/templates/stateful-set.yaml
mpereira: Yes! I could swear I had created the JIRA for that, but apparently not. Just created https://jira.mesosphere.com/browse/DCOS-60288 for that. All Docker images and their pull policies should be configurable through parameters.

mpereira:
Also, I think it's ok that for this PR to not make `cassandra_exporter` totally configurable through `params.yaml`. I'll create a separate ticket for that.

mpereira:
Created https://jira.mesosphere.com/browse/DCOS-60293 for the above.

operator/templates/stateful-set.yaml
samvantran: This sounds kind of involved - why don't we make this a separate ticket for now so that we don't block the metrics work. If we're okay with personal images for now, let's continue using them until the Docker image build support is in place.

resources/service-monitor.yaml
viivek46: As we are using kubectl create, it wont take param like that.
Also "default" is namespace of Cassandra cluster not Prometheus.

operator/params.yaml
```diff
@@ -22,6 +22,22 @@ NODE_MEM_LIMIT_MIB:
   description: "Memory limit (in MiB) for the Cassandra node pods."
   default: 4096
 
+METRICS_CPU_MC:
```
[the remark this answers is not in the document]
viivek46: @mpereira Okay i will prefix it as suggested , and for time being its okay. but I used METRICS for  general purpose as Kudo Cassandra is exporting metrics which could be used for other tools, and if customer wants to use Prometheus we are specifically running this command for it  `kubectl create -f resources/service-monitor.yaml` , thats why `service-monitor.yaml` was not included in templating (hence cant provide param to service-monitor.yaml).

cc: @samvantran

operator/templates/stateful-set.yaml
viivek46: @mpereira we currently need those capabilities to run our commands.

mpereira: Which commands?

viivek46: https://github.com/mesosphere/kudo-cassandra-operator/blob/metrics/operator/templates/stateful-set.yaml#L157
this is error we get if we dont use security context : 
`bash: /usr/bin/java: Operation not permitted`

mpereira: That's because the prometheus exporter is using the Cassandra image. It should use its own image: https://github.com/mesosphere/kudo-cassandra-operator/pull/4#discussion_r337509622

resources/service-monitor.yaml
mpereira: If the `default` here is related to the Cassandra operator namespace, it should be `{{ .Namespace }}` then. I was under the impression that it was the namespace for the Prometheus instance.

And we can remove `SERVICE_MONITOR_PROMETHEUS_NAMESPACE` too.

operator/params.yaml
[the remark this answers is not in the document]
mpereira: Regarding requiring a manual `kubectl create -f resources/service-monitor.yaml` command, can we just enclose the whole `resources/service-monitor.yaml` in an "if prometheus exporter enabled" block instead?

operator/templates/stateful-set.yaml
mpereira: Yeah, step 3 above calls out that the image can be personal.

images/Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM cassandra:3.11.4
 RUN set -eux; \
   apt-get update; \
   apt-get install -y --no-install-recommends libcap2-bin; \
+  apt-get install -y curl; \
```
mpereira: Let's revert this.

resources/service-monitor.yaml
viivek46: yea my bad forgot to remove SERVICE_MONITOR_PROMETHEUS_NAMESPACE.

operator/templates/stateful-set.yaml
viivek46: Correct, with new images it wont be required, but currently it was required in cassandra image,  but i didnt updated this PR yet because of : https://github.com/mesosphere/kudo-cassandra-operator/pull/4#discussion_r337733759

operator/params.yaml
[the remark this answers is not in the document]
viivek46: @mpereira @samvantran 
Although i have already changed Param to prefix with PROMETHEUS_EXPORTER, i just want to confirm again that should we be targeting only Prometheus ? or just make it more general and ask customer to run command manually to see metrics on Prometheus. 
if we are targeting Prometheus only then this service-monitor object can be automated can be automated.Currently  `resources/service-monitor.yaml` is not  included in operator hence its manual process and at the same time we cant use any param in that. 

also just FYI General approach is followed by kudo kafka.

resources/service-monitor.yaml
viivek46: also 
https://github.com/mesosphere/kudo-cassandra-operator/pull/4#discussion_r338155501

operator/params.yaml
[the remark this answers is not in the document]
mpereira: Yes, let's focus on Prometheus.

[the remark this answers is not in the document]
samvantran: I think for beta, it's fine for us to specifically target prometheus since we're also using the criteo/prom-exporter. But when we GA, if we decide to output custom/native metrics then we can generalize some of these configs

operator/templates/cassandra-exporter-config-yml.yaml
samvantran: Nice - good addition. I think someone in there probably just a ctrl+find+all and accidentally typo'd

operator/templates/service.yaml
```diff
@@ -2,6 +2,10 @@ apiVersion: v1
 kind: Service
 metadata:
   name: svc
+  {{ if eq .Params.PROMETHEUS_EXPORTER_ENABLED "true" }}
+  labels:
+    "kudo.dev/servicemonitor": "true"
```
mpereira:
```suggestion
    kudo.dev/servicemonitor: "true"
```

operator/templates/stateful-set.yaml
mpereira: Please revert line 49 for now @viivek46.

docs/README.md
```diff
@@ -0,0 +1,4 @@
+## KUDO Cassandra Operator
+
+- [Monitoring](./monitoring.md)
+
```
mpereira:
```suggestion
```

docs/monitoring.md
```diff
@@ -0,0 +1,25 @@
...
+Port:              prometheus-exporter-port  7200/TCP
+TargetPort:        7200/TCP
+...
+```
+
+If the Cassandra Service in the default namespace, we will need to use the next example of the service-monitor to see metrics in Prometheus.
+
+```
+kubectl create -f resources/service-monitor.yaml
+```
```
mpereira: We should add instructions to how to import the Grafana dashboard. And probably remove what's currently written, since the service monitor will be installed by default.

viivek46: Correct with new changes which include automation I will update this instructions, i just pushed code as "wip" as we were discussing about an issue.

operator/templates/stateful-set.yaml
viivek46: we have to modify exporter docker file now, as cassandra_exporter.jar file was not accessible by non root user, i think we should skip step 1,2 from Murilo's suggestion and copy paste run.sh and docker file from crito's repo and modify them in our repo.

WDYT @mpereira @samvantran

resources/service-monitor.yaml
viivek46: service-monitor.yaml is automated now as we are focusing on Prometheus only.

images/Dockerfile.prometheus-exporter
```diff
@@ -0,0 +1,3 @@
+FROM criteord/cassandra_exporter:2.2.1
+
+RUN chown 999:999 /opt/cassandra_exporter/cassandra_exporter.jar
```
porridge: @viivek46 can you please explain why we need to `chown` the file? What runs as UID 999 and why does it need to own this JAR?

mpereira: > @porridge: can you please explain why we need to chown the file? What runs as UID 999 and why does it need to own this JAR?

Because of the [pod security context](https://github.com/mesosphere/kudo-cassandra-operator/blob/ea8152a05e76fadc525bfe94f85445dee76b6006/operator/templates/stateful-set.yaml#L24-L36) this image container runs under.

porridge: Thanks @mpereira . However `chown`-ing the file gives the process the permission to modify the code to be run, which is frowned upon from security PoV.
I'm guessing the intent probably to `chmod a+r` it...
````
-rw-rwx--- 1 root root 4607224 Jan 30  2019 cassandra_exporter.jar
```
Right. So the fix should actually be done upstream, let me take a stab at it.

porridge: https://github.com/criteo/cassandra_exporter/pull/68

PR #6, opened 2019-10-23
Implement Readiness Probe

phrynchanka:
## What changes are proposed in this PR? 
Resolves  [DCOS-58972](https://jira.mesosphere.com/browse/DCOS-58972) 

This PR Implement Readiness Probe for Kudo Cassandra operator. 
LivenessProbe wasn't implemented because we need clear understand when pods definitely must be [restarted](https://cloud.google.com/blog/products/gcp/kubernetes-best-practices-setting-up-health-checks-with-readiness-and-liveness-probes).

Implementation steps : 
1. Added k8s ReadinessProbe properties to `params.yml`
2. Modified `stateful-set.yaml`
3. Added ReadinessProbe script similar like in [`DC/OS`](https://github.com/mesosphere/dcos-commons/blob/master/frameworks/cassandra/src/main/dist/svc.yml#L154-L168) but without security support (which should be implemented and script modified) 
## How were these changes tested?
Manually tested  
1. Install operator 
2. Look into pod events 
`kubectl get event --namespace $kudo_cassandra_instance_namespace --field-selector involvedObject.name=cassandra-node-0` 
and
`kubectl describe pod cassandra-node-0`
3. Check that all pods were run and healthy.


operator/templates/stateful-set.yaml
```diff
@@ -62,6 +62,13 @@ spec:
                 command:
                   - nodetool
                   - drain
+          readinessProbe:
+            exec:
+              command:
+                - nodetool status -p {{ .Params.JMX_PORT }}
```
phrynchanka: @mpereira 
What is the cassandra working dir? Prob failed (I tried different ways to run)
```
Readiness probe errored: rpc error: code = Unknown desc = failed to exec in container: failed to start exec "8ffdd0483521cc3b6dad2b461179285023ba852179c50c919dec6c2f915b2f6a": OCI runtime exec failed: exec failed: container_linux.go:345: starting container process caused "exec: \"/var/lib/cassandra/nodetool status -p 7199\": stat /var/lib/cassandra/nodetool status -p 7199: no such file or directory": unknown
```

operator/templates/cassandra-healthcheck-sh.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cassandra-healthcheck-sh
+data:
+  cassandra-healthcheck.sh: |
+    echo "Hello!"
```
phrynchanka: This script should be like here https://github.com/mesosphere/dcos-commons/blob/master/frameworks/cassandra/src/main/dist/svc.yml#L154-L168
But on readiness check phase ConfigMap is not available.
@mpereira  Do you have some ideas how to get around it?

mpereira: What's the evidence that it wasn't available?

phrynchanka: resolved

operator/params.yaml
```diff
@@ -701,3 +701,23 @@ JVM_OPT_MAX_GC_PAUSE_MILLIS:
 JVM_OPT_G1R_SET_UPDATING_PAUSE_TIME_PERCENT:
   description: "Have the JVM do less remembered set work during STW, instead preferring concurrent GC. Reduces p99.9 latency."
   default:
+
+READINESS_CHECK_INITIAL_DELAY:
+  description: "An amount of time in seconds to wait before starting the readiness check attempts."
+  default: "0"
+
+READINESS_CHECK_PERIOD:
+  description: "The period in seconds to wait after the last readiness check has completed to start the next check."
```
samvantran:
```suggestion
  description: "How often (in seconds) to perform the readiness check."
```

operator/params.yaml
```diff
@@ -701,3 +701,23 @@ JVM_OPT_MAX_GC_PAUSE_MILLIS:
...
+  description: "An amount of time in seconds to wait for a readiness check to succeed."
+  default: "60"
+
+READINESS_SUCCESS_THRESHOLD:
+  description: "Minimum consecutive successes for the probe to be considered successful after having failed."
+  default: "1"
+
+READINESS_FAILURE_THRESHOLD:
+  description: "When a Pod starts and the probe fails, Kubernetes will try failureThreshold times before giving up. Giving up means marking pod Unready."
+  default: "3"
```
samvantran:
```suggestion
  default: "3"

```
added a new line

operator/params.yaml
```diff
@@ -701,3 +701,23 @@ JVM_OPT_MAX_GC_PAUSE_MILLIS:
...
+READINESS_CHECK_TIMEOUT:
+  description: "An amount of time in seconds to wait for a readiness check to succeed."
+  default: "60"
+
+READINESS_SUCCESS_THRESHOLD:
+  description: "Minimum consecutive successes for the probe to be considered successful after having failed."
+  default: "1"
+
+READINESS_FAILURE_THRESHOLD:
+  description: "When a Pod starts and the probe fails, Kubernetes will try failureThreshold times before giving up. Giving up means marking pod Unready."
```
samvantran:
```suggestion
  description: "When a Pod starts and the check fails, Kubernetes will try failure_threshold times before giving up. Giving up means marking the pod Unready."
```

operator/params.yaml
```diff
@@ -701,3 +701,23 @@ JVM_OPT_MAX_GC_PAUSE_MILLIS:
 JVM_OPT_G1R_SET_UPDATING_PAUSE_TIME_PERCENT:
   description: "Have the JVM do less remembered set work during STW, instead preferring concurrent GC. Reduces p99.9 latency."
   default:
+
+READINESS_CHECK_INITIAL_DELAY:
+  description: "An amount of time in seconds to wait before starting the readiness check attempts."
```
samvantran: Let's use the [k8s terminology](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes)

```suggestion
  description: "Number of seconds after the container has started before the readiness check is initiated."
```

operator/params.yaml
```diff
@@ -701,3 +701,23 @@ JVM_OPT_MAX_GC_PAUSE_MILLIS:
...
+
+READINESS_CHECK_PERIOD:
+  description: "The period in seconds to wait after the last readiness check has completed to start the next check."
+  default: "5"
+
+READINESS_CHECK_TIMEOUT:
+  description: "An amount of time in seconds to wait for a readiness check to succeed."
+  default: "60"
+
+READINESS_SUCCESS_THRESHOLD:
```
samvantran: The other params are prefixed with `READINESS_CHECK`. Can you do the same for the others or drop `CHECK` from the first two params?

operator/params.yaml
```diff
@@ -701,3 +701,23 @@ JVM_OPT_MAX_GC_PAUSE_MILLIS:
...
+  description: "An amount of time in seconds to wait before starting the readiness check attempts."
+  default: "0"
+
+READINESS_CHECK_PERIOD:
+  description: "The period in seconds to wait after the last readiness check has completed to start the next check."
+  default: "5"
+
+READINESS_CHECK_TIMEOUT:
+  description: "An amount of time in seconds to wait for a readiness check to succeed."
+  default: "60"
```
samvantran: do these have to be strings? can they be ints?

operator/templates/stateful-set.yaml
```diff
@@ -62,6 +62,16 @@ spec:
...
                   - drain
+          readinessProbe:
+            exec:
+              command:
+                  - /bin/bash
+                  - /etc/cassandra/cassandra-readiness-probe.sh
+            initialDelaySeconds: {{ .Params.READINESS_CHECK_INITIAL_DELAY }}
+            periodSeconds: {{ .Params.READINESS_CHECK_PERIOD }}
+            timeoutSeconds: {{ .Params.READINESS_CHECK_TIMEOUT }}
+            successThreshold: {{ .Params.READINESS_SUCCESS_THRESHOLD }}
```
samvantran: I wonder if this param is even needed to be configurable since we only need 1 successful check. But in the spirit of configuring :all-the-things:...

operator/operator.yaml
```diff
@@ -16,6 +16,7 @@ tasks:
       - cassandra-env-sh.yaml
       - jvm-options.yaml
       - stateful-set.yaml
+      - cassandra-readiness-probe-sh.yaml
```
samvantran: You're using `readiness_probe` here and `readiness_check` in other places. Let's use one or the other? Since we're in the world of k8s, let's go with `probe`, WDYT?

phrynchanka: > You're using `readiness_probe` here and `readiness_check` in other places. Let's use one or the other? Since we're in the world of k8s, let's go with `probe`, WDYT?

Make sense. Thanks!

operator/templates/stateful-set.yaml
phrynchanka: I've exposed all properties which k8s offer for readiness/liveness probe. Maybe it will be used :)

phrynchanka:
> Also, I noticed that a slightly different pattern is [used elsewhere](https://github.com/bitnami/charts/blob/master/bitnami/cassandra/templates/statefulset.yaml#L166) for grepping the `nodetool` output. Did it also work with the pattern we're using here?

Yep, our pattern works great.

operator/params.yaml
phrynchanka: done. thanks!

phrynchanka:
> Let's wait for the tests to be merged to master and then we'll get this one merged as well.
Thank you!
Do we need to implement integration test?

mpereira:
I think for now the existing test should be sufficient, since a successful install is dependent on the readiness probes. We might want to add more specific tests for it in the near future, but for now it should be ok.

operator/params.yaml
```diff
@@ -30,6 +30,26 @@ NODE_STORAGE_CLASS:
   description: "The storage class to be used in volumeClaimTemplates. By default, it is not required and the default storage class is used."
   required: false
 
+NODE_READINESS_PROBE_INITIAL_DELAY_S:
```
samvantran: Is the `_S` necessary if the description is already clear? Same question for period and timeout.

mpereira:
🚢

PR #13, opened 2019-10-30
Parameterize Prometheus Exporter Docker image.

mpereira:


operator/templates/stateful-set.yaml
```diff
@@ -150,8 +150,8 @@ spec:
               subPath: node-readiness-probe.sh
         {{ if eq .Params.PROMETHEUS_EXPORTER_ENABLED "true" }}
         - name: prometheus-exporter
-          imagePullPolicy: Always
-          image: mesosphere/cassandra-prometheus-exporter:0.1.0-2.2.1-SNAPSHOT
+          image: {{ PROMETHEUS_EXPORTER_DOCKER_IMAGE }}
```
viivek46:
```suggestion
          image: {{ .Params.PROMETHEUS_EXPORTER_DOCKER_IMAGE }}
```

operator/templates/stateful-set.yaml
```diff
@@ -150,8 +150,8 @@ spec:
...
+          imagePullPolicy: {{ PROMETHEUS_EXPORTER_DOCKER_IMAGE_PULL_POLICY }}
```
viivek46:
```suggestion
          imagePullPolicy: {{ .Params.PROMETHEUS_EXPORTER_DOCKER_IMAGE_PULL_POLICY }}
```

mpereira:
> why we are exposing Prometheus image in param.

We want to make it configurable at the operator level.

> also as we have made changes here should we revert back changes from '/templates' and metadata ?

I'll update the Docker image metadata handling on a separate PR. Maybe we'll make `params.yaml` a template.

mpereira:
Thanks for fixing the parameter accessors!

PR #23, opened 2019-11-11
Create a script that automatically creates a markdown table with all parameters

phrynchanka:


mpereira:
Thanks @phrynchanka! I tried it out locally and it works perfectly! I'm gonna push a few improvements in a follow up PR.

PR #25, opened 2019-11-13
Fix Grafana dashboard for Prometheus metrics.

phrynchanka:
Fixed Grafana json from https://github.com/criteo/cassandra_exporter/blob/master/grafana/cassandra_kubernetes.json

### DIFF
--- a/monitoring/grafana/cassandra.json
+++ b/monitoring/grafana/cassandra.json
@@ -1,0 +1,4110 @@
+{
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS",
+      "label": "prometheus",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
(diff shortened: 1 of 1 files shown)


mpereira:
Thanks @phrynchanka.

Would you be able to attach a screenshot of the dashboard's graphs working?

phrynchanka:
> Thanks @phrynchanka.
> 
> Would you be able to attach a screenshot of the dashboard's graphs working?

Sure. I've run stress test to see graphs.
---
cassandra_grafana

mpereira:
Thanks @phrynchanka, that looks better.

I wonder if the graphs that show up empty are like that because there's really no metrics for them, or because their query is incorrect. Do you have an idea?

mpereira:
I'm gonna merge this for now so that we have the dashboard in the repository. We can work on investigating why some graphs are empty later.

PR #22, opened 2019-11-11
Add operator upgrade integration test

viivek46:
## What changes are proposed in this PR? 
Implement tests for [DCOS-58981](https://jira.mesosphere.com/browse/DCOS-59496)

This PR include :
- Add new upgrade suit 
- upgrading operator from tarball URL

PR #17, opened 2019-11-04
upgrade cassandra to 3.11.5

viivek46:
## What changes are proposed in this PR? 
Resolves  [DCOS-58981](https://jira.mesosphere.com/browse/DCOS-58981)

## How were these changes tested?
CI tests to check functionality of cassandra-3.11.5
Upgrade from 3.11.4 to 3.11.5 is manually tested and Below are steps to check :
1. Install Cassandra 3.11.4 operator, check deploy plan status to `COMPLETE`
2. Run below upgrade command from cassandra 3.11.5 operator folder : 
```
kubectl kudo upgrade ./operator         --instance="${kudo_cassandra_instance_name}"         --namespace="${kudo_cassandra_instance_namespace}"
```
3. To check operator verions we can run  
`kubectl  get operatorversions`

PR #29, opened 2019-12-10
Upgrade C* to 3.11.5, add integration tests

mpereira:
- 2dc8c4d: Make `params.yaml` a template (it should have been one already)
- 821ce5f: Upgrade C* to 3.11.5
- 51521cb: Add integration test for upgrade
- 2ffc70c: Create function to return a C* cluster nodes
- ec47331: Assert the C* has the expected number of UN nodes all around

This supersedes #17 and #22.


mpereira:
To do: add "Unreleased" entry to the [CHANGELOG](https://github.com/mesosphere/kudo-cassandra-operator/blob/master/CHANGELOG.md).

operator/templates/cassandra-yaml.yaml
```diff
@@ -558,6 +558,21 @@ data:
...
+    #
+    # The default setting of 18 generates trees of maximum size around
+    # 50 MiB / tree. If you are running out of memory during repairs consider
+    # lowering this to 15 (~6 MiB / tree) or lower, but try not to lower it
+    # too much past that or you will lose too much resolution and stream
+    # too much redundant data during repair. Cannot be set lower than 10.
+    #
+    # For more details see https://issues.apache.org/jira/browse/CASSANDRA-14096.
+    #
+    {{ if .Params.REPAIR_SESSION_MAX_TREE_DEPTH }}
```
zmalik: I think KUDO treats empty string and nil different. So this might be true with current defaults. Setting no default value is different than `""` default value. Same for new params introduced in this PR with default value `""`

mpereira: For future reference: it seems like at least as of right now (KUDO 0.8.0), a template with `if ""` will not trigger the if-condition.

`params.yaml`:
```
  - name: ENABLE_SASI_INDEXES
    description: "Enables SASI index creation on this node. SASI indexes are considered experimental and are not recommended for production use."
    default: ""
```

`cassandra.yaml`:
```
    # Enables SASI index creation on this node.
    # SASI indexes are considered experimental and are not recommended for production use.
    {{ if .Params.ENABLE_SASI_INDEXES }}
    enable_sasi_indexes: {{ .Params.ENABLE_SASI_INDEXES }}
    {{ end }}
```

The line with `enable_sasi_indexes` is not present in the file compiled by KUDO:
```
$ kubectl exec "pod/${kudo_cassandra_instance_name}-${kudo_cassandra_statefulset_name}-${pod}" \
          -n "${kudo_cassandra_instance_namespace}" \
          -c "cassandra" \
          -- \
          bash -c '
            cat "/etc/cassandra/cassandra.yaml" | grep -i sasi
          '
# Enables SASI index creation on this node.
# SASI indexes are considered experimental and are not recommended for production use.
```

PR #31, opened 2020-01-13
Add support to enable transport encryption TLS

shubhanilBag:
This PR:

1. Adds configurable params to enable transport encryption
2. Supports adding custom TLS CA cert using a configurable k8s secret
3. Adds a clqshrc ConfigMap file that allows cqlsh to be used without any additional configuration when TLS is enabled
4. Adds integration tests to check
  1.1 Node-to-Node transport encryption
  1.2 Client-to-Node encryption
  1.3 Node-to-Node along with Client-to-Node encryption
  1.4 Data read and write using cqlsh
5. Adds and updates documentation for enabling TLS
~~6. Bumps operator version to `0.2.0`~~


operator/params.yaml
```diff
@@ -123,6 +123,42 @@ parameters:
...
+  - name: TRANSPORT_ENCRYPTION_CIPHERS
+    description: "Comma-separated list of JSSE Cipher Suite Names."
+    default: "TLS_RSA_WITH_AES_128_CBC_SHA,TLS_RSA_WITH_AES_256_CBC_SHA,TLS_DHE_RSA_WITH_AES_128_CBC_SHA,TLS_DHE_RSA_WITH_AES_256_CBC_SHA,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA"
+
+  - name: TRANSPORT_ENCRYPTION_CLIENT_ALLOW_PLAINTEXT
+    description: "Enable Server-Client plaintext communication alongside encrypted traffic."
+    default: "false"
+
+  - name: TRANSPORT_ENCRYPTION_REQUIRE_CLIENT_AUTH
+    description: "TODO"
```
zmalik: lets add some description here

mpereira:
@shubhanilBag could you please merge in the latest master?

operator/operator.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: kudo.dev/v1beta1
 name: "cassandra"
-operatorVersion: "0.1.2"
+operatorVersion: "0.2.0"
```
mpereira: Let's revert the operator version bump. Version bumps should happen in the stable branches.

operator/params.yaml
mpereira: +1, and the following ones as well.

operator/templates/generate-cqlshrc-sh.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Name }}-generate-cqlshrc-sh
+  namespace: {{ .Namespace }}
+data:
+  generate-cqlshrc.sh: |
```
mpereira: Could this be a configmap of `cqlshrc` itself instead of a bash script that generates it?

templates/operator/templates/stateful-set.yaml.template
```diff
@@ -148,6 +154,17 @@ spec:
             - name: node-readiness-probe-sh
               mountPath: /etc/cassandra/node-readiness-probe.sh
               subPath: node-readiness-probe.sh
+            - name: cassandra-home
+              mountPath: /home/cassandra/.cassandra/
```
mpereira: Should `cassandra-home` be just `/home/cassandra/`?

tests/utils/k8s/k8s.go
```diff
@@ -128,3 +128,25 @@ func ExecInPodContainer(
 
 	return stdout, nil
 }
+
+func FetchLogsOfPod(namespaceName, podName, containerName string) (*bytes.Buffer, error) {
```
mpereira: Would `PodContainerLogs` be a more representative name for this function?

tests/utils/cassandra/cassandra.go
```diff
@@ -120,3 +122,19 @@ func getConfigurationFromNodeLogs(
 
 	return configuration, nil
 }
+
+// CQLSH TODO function comment.
+func CQLSH(
```
mpereira:
```suggestion
func cqlsh(
```

tests/suites/cassandra_tls/tls_test.go
```diff
@@ -0,0 +1,214 @@
...
+			if err != nil {
+				Fail(
+					"Failing the full suite: failed to install operator instance that the " +
+						"following tests depend on",
+				)
+			}
+			Expect(err).To(BeNil())
+			assertNumberOfCassandraNodes(NodeCount)
+		})
+		It("Check container logs", func() {
```
mpereira: Try to follow the active voice for the test names, so that it reads like:
```
It "checks for the container logs"
```
like the above
```
It "installs the operator from a directory"
```
On the other tests as well.

tests/suites/cassandra_tls/tls_test.go
```diff
@@ -0,0 +1,214 @@
+package cassandra_tls
```
mpereira: Let's put this test file in `tests/suites/tls_test.go`.

tests/utils/k8s/k8s.go
shubhanilBag: Let's go with `FetchContainerLogs`

templates/operator/templates/stateful-set.yaml.template
shubhanilBag: Both cqlsh and nodetool creates files in `.cassandra`, /home/cassandra is not used directly
so directly mounted it there, otherwise it would need another mkdir.

> cassandra@cassandra-node-0:~ $ ls -lah
total 0
drwxr-xr-x. 3 root root      24 Jan 29 14:46 .
drwxr-xr-x. 1 root root      23 Jan 29 14:46 ..
drwxrwsrwx. 2 root cassandra 66 Jan 29 14:49 .cassandra
cassandra@cassandra-node-0:~$ ls -lah .cassandra/
total 12K
drwxrwsrwx. 2 root      cassandra   66 Jan 29 14:49 .
drwxr-xr-x. 3 root      root        24 Jan 29 14:46 ..
-rw-------. 1 cassandra cassandra    5 Jan 29 14:49 cqlsh_history
-rw-r--r--. 1 cassandra cassandra  264 Jan 29 14:46 cqlshrc
-rw-r--r--. 1 cassandra cassandra 1.7K Jan 29 14:50 nodetool.history

mpereira: Let's name it something other than "cassandra-home" then, since it implies `/home/cassandra`. Maybe something like `cassandra-configuration-directory` or `dot-cassandra`.

operator/templates/generate-cqlshrc-sh.yaml
shubhanilBag: That is what I was going to but, I made a script so that in future if additional configuration is needed for cqlsh, it can be in one script; like adding custom config containing additional config for cqlsh ref: https://docs.datastax.com/en/cql-oss/3.3/cql/cql_reference/cqlshUsingCqlshrc.html

tests/utils/k8s/k8s.go
mpereira: How is that function different than the [already existing `k8s.GetPodContainerLogs`](https://github.com/mesosphere/kudo-cassandra-operator/blob/8932a7e5f58b52495466ca16265aace4b8b3ca97/tests/utils/k8s/k8s.go#L75) by the way?

templates/operator/templates/stateful-set.yaml.template
shubhanilBag: `dot-cassandra` SGTM :+1:

tests/utils/k8s/k8s.go
shubhanilBag: It's not,  I made a duplicate by mistake :thinking:! Thanks for pointing it out, will fix that

operator/templates/generate-cqlshrc-sh.yaml
mpereira: What's the difference in having a templated bash script that outputs a file from a templated file?

tests/suites/cassandra_tls/tls_test.go
shubhanilBag: I followed kudo-kafka-operator's way of having tests in different packages with it's own folder. If we have the `tls_test.go` and `sanity_test.go` under same package sanity they will share the global vars and functions. We won't get full isolation. like https://github.com/mesosphere/kudo-cassandra-operator/blob/master/tests/suites/sanity_test.go#L22 and https://github.com/mesosphere/kudo-cassandra-operator/blob/8932a7e5f58b52495466ca16265aace4b8b3ca97/tests/suites/cassandra_tls/tls_test.go#L19 will conflict

mpereira: I see. In that case, let's move
- `tests/suites/sanity_test.go` to `tests/suites/sanity/sanity_test.go`
- `tests/suites/cassandra_tls/tls_test.go` to `tests/suites/tls/tls_test.go`.

shubhanilBag: Sure :+1:

operator/templates/generate-cqlshrc-sh.yaml
shubhanilBag: like here what we do for kafka custom properties taken from a user provided CM https://github.com/kudobuilder/operators/blob/ebaf9e6dba463aa4ff8ec7b4e537c7fdc8c3f629/repository/kafka/operator/templates/bootstrap.yaml#L175

shubhanilBag: If this is not required for cqlshrc file then I can revert back to the usual way for mounting a CM and cp it

tests/utils/kudo/kudo.go
```diff
@@ -411,7 +411,7 @@ func UpgradeOperator(
 func UninstallOperator(
 	operatorName string, namespaceName string, instanceName string,
 ) error {
-	uninstallScript := "../../scripts/uninstall_operator.sh"
+	uninstallScript := "../../../scripts/uninstall_operator.sh"
```
ANeumann82: This is required because the suites moved one directory deeper? 

Maybe we should pass in the scripts directory as an absolute path at some point and not rely on relative paths. Seems to be ok for now though.

mpereira: Would be ok for now for me too. Is there a way to make this relative to the actual `kudo.go` file instead, though?

tests/suites/cassandra_sanity/sanity_test.go
```diff
@@ -49,7 +49,12 @@ var _ = Describe(TestName, func() {
 	It("Installs the latest operator from the package registry", func() {
 		// TODO(mpereira) Assert that it isn't running.
 		err := kudo.InstallOperator(
-			OperatorName, TestNamespace, TestInstance, []string{},
+			OperatorName, TestNamespace, TestInstance, []string{
+				"NODE_CPU_MC=200",
+				"NODE_MEM_MIB=800",
+				"PROMETHEUS_EXPORTER_CPU_MC=100",
+				"PROMETHEUS_EXPORTER_MEM_MIB=200",
```
ANeumann82: These are probably development settings for local testing?

tests/utils/kudo/kudo.go
shubhanilBag: @ANeumann82 right, later we can do something like this: https://github.com/mesosphere/kudo-kafka-operator/pull/6/files#diff-dc3495b5db55989eec09ceab2776f370R29
Set a `REPO_ROOT` env var and create abs paths using that

mpereira: It's already available to be passed as an environment variable fwiw: https://github.com/mesosphere/kudo-cassandra-operator/blob/7a143804c906b3f8c5fda40f359ab297506c35ff/run-tests.sh#L7

tests/suites/cassandra_sanity/sanity_test.go
shubhanilBag: right! I will remove these :ballot_box_with_check: from here and tls test

shubhanilBag: but ts a good practice to have tests that consume as less resource as possible, I added these as the default ones where too large to run in konovy cluster with default config.
It can help in starting up running tests in small clusters faster, plus if we have extra pods in tests like KDC or like a client pod for testing out application features then those can fail to run due to resource starvation on the k8s nodes WDYT @ANeumann82 ?

ANeumann82: Yeah, I'm fine with that. I didn't know Cassandra enough yet to tune the settings down, but as long as we don't make anything flaky because Cassandra doesn't have enough CPU or Mem, i'm fine with having lower values for the tests.

mpereira: These settings are too low, and can likely cause resource throttling issues. I've seen it happen during MWTs. Unless we're specifically testing CPU/MEM parameters, let's use the default parameter values.

operator/templates/generate-cqlshrc-sh.yaml
shubhanilBag: @mpereira as discussed we are keeping the script based approach as POD_NAME env variable is used in the script to create the hostname

shubhanilBag:
@zmalik @mpereira @ANeumann82 thanks for the thorough review :+1: added changes as requested in https://github.com/mesosphere/kudo-cassandra-operator/pull/31/commits/fd0ff10da20e116e70e28ac4c8034b43d94d8fdc

README.md
```diff
@@ -48,6 +48,7 @@ kubectl kudo install cassandra
 - [Managing](/docs/managing.md)
 - [Upgrading](/docs/upgrading.md)
 - [Monitoring](/docs/monitoring.md)
+- [Security](./docs/secuity.md)
```
mpereira:
```suggestion
- [Security](./docs/security.md)
```

docs/security.md
```diff
@@ -0,0 +1,68 @@
+# Securing KUDO Cassandra Operator instances
+
+The KUDO Cassandra service supports Cassandra’s native transport **encryption**
```
mpereira: Let's rephrase all instances of "service" with "operator" to maintain consistency.
```suggestion
The KUDO Cassandra operator supports Cassandra’s native transport **encryption**
```

docs/security.md
```diff
@@ -0,0 +1,68 @@
+# Securing KUDO Cassandra Operator instances
+
+The KUDO Cassandra service supports Cassandra’s native transport **encryption**
+mechanism. The service provides automation and orchestration to simplify the use
+of these important features. For more information on Cassandra’s security, read
```
mpereira: Let's also rephrase "Cassadra" instances to "Apache Cassandra" for the same reason.
```suggestion
of these important features. For more information on Cassandra’s security, read

```

docs/security.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+:warning: Make sure to create the certificate in the same namespace where the
+KUDO Cassandra is being installed.
+
+#### Enabling only Node-to-node communication
+
+```
+kubectl kudo install cassandra \
+    --instance=cassandra --namespace=kudo-cassandra \
```
mpereira: Let's maintain consistency of either all parameters in one line, or one parameter per-line when there's lots of them.

```suggestion
    --instance=cassandra \
    --namespace=kudo-cassandra \
```

mpereira:
I pushed two small commits fixing a documentation thing.

@shubhanilBag would you be able to add an ["unreleased" changelog entry](https://github.com/mesosphere/kudo-cassandra-operator/blob/master/DEVELOPMENT.md#additional-release-work) as well?

shubhanilBag:
> I pushed two small commits fixing a documentation thing.
> 
> @shubhanilBag would you be able to add an ["unreleased" changelog entry](https://github.com/mesosphere/kudo-cassandra-operator/blob/master/DEVELOPMENT.md#additional-release-work) as well?

Added the doc fixes and added changelog entry in https://github.com/mesosphere/kudo-cassandra-operator/pull/31/commits/b3bf8f8eab74e8bdc672130eb8139633b9aebded

mpereira:
🚢

mpereira:
Looks like there are two test failures:
```
[18:14:38][Step 5/7] Summarizing 2 Failures:
[18:14:38][Step 5/7] 
[18:14:38][Step 5/7] [Fail] sanity-test [It] Configures Cassandra properties through custom properties 
[18:14:38][Step 5/7] /kudo-cassandra-operator/tests/suites/sanity/sanity_test.go:151
[18:14:38][Step 5/7] 
[18:14:38][Step 5/7] [Fail] sanity-test [It] Configures Cassandra JVM options through custom options 
[18:14:38][Step 5/7] /kudo-cassandra-operator/tests/suites/sanity/sanity_test.go:175
[18:14:38][Step 5/7] 
[18:14:38][Step 5/7] Ran 7 of 7 Specs in 1109.357 seconds
[18:14:38][Step 5/7] FAIL! -- 5 Passed | 2 Failed | 0 Pending | 0 Skipped
```
@shubhanilBag could you take a look when you get a chance?

shubhanilBag:
The sanity test suite was running the tests in this order:

1. install from community repo
2. upgrade to local version
3. Node Scaling tests (3nodes -> 4nodes)
4. Three parameter update tests

As the result the 3 param update test were running with 4 nodes in the c* instance, so the SS took more time than the estimated 5mins of timeout and failing the tests
In https://github.com/mesosphere/kudo-cassandra-operator/pull/31/commits/c0fb3bddb918462659f96df621a102245836ea02 I moved the node scaling tests to the end. The tests are passing and the update tests are a bit faster now.
cc: @mpereira

tests/utils/cassandra/cassandra.go
```diff
@@ -120,3 +122,19 @@ func getConfigurationFromNodeLogs(
 
 	return configuration, nil
 }
+
+// Cqlsh TODO function comment.
```
zmalik: add the comment?

shubhanilBag: added in https://github.com/mesosphere/kudo-cassandra-operator/pull/31/commits/a2162f4231fbebb5b27abc09ffebed78bbc0b93c

PR #37, opened 2020-01-17
Link to upgrading doc too

porridge:


porridge:
BTW these links are all broken when viewing this file through the symlink in `docs`..

mpereira:
Hm, good find. Maybe we can use non-relative paths there somehow?

PR #41, opened 2020-01-20
Operator snapshot version with date, git sha and dirtyness information.

mpereira:
This will help with reproducibility and traceability of issues in snapshot builds (e.g., CI and development).

Notice that the UTC ISO 8601 timestamp is using lowercase "T" and "Z" characters. This is because of a [bug in KUDO](https://github.com/kudobuilder/kudo/issues/1294) which should be addressed shortly.

The snapshot format is documented in this [DEVELOPMENT.md section](https://github.com/mesosphere/kudo-cassandra-operator/blob/ed9da98fa46e1b3136e5036c0e5f553d847b8dc3/DEVELOPMENT.md#snapshots).


mpereira:
This will also fix the current situation with the upgrade integration test failing due to the source and target operator versions being the same. With this, the current test will install KUDO C* from the registry (e.g. version `3.11.5-0.2.1`) and will upgrade to a snapshot version (e.g. `3.11.5-0.2.1-20200120t170941z-2c1fe686bede`) .

mpereira:
This will also make releasing non-stable versions or dirty git repositories *very* visible (assuming that CI isn't checking for these things yet).

DEVELOPMENT.md
```diff
@@ -405,18 +405,18 @@ further steps are required.
...
 
-### Snapshots (to be implemented)
+### Snapshots
 
-Snapshot release versions will have the following format:
+Snapshot release versions have the following format:
 
 ```
-${app_version}-0.0.0-YYYYMMDDHHmmss-${git_sha}
+${app_version}-${operator_version}-YYYYMMDDTHHmmssZ-${git_sha}
```
porridge: So is it upper or lower case T and Z?

mpereira:
@porridge you're right. This would be better handled by an explicit build process. Closing this for now.

PR #43, opened 2020-01-21
Overwrite operatorVersion on upgrade tests.

mpereira:
This fixes the upgrade test where

``` bash
kudo install cassandra --instance cass01
kudo upgrade ./operator --instance cass01
```

fails because the `operatorVersion` in `./operator` is the same as the one in the `cassandra` package available from the registry.


mpereira:
More context: https://mesosphere.slack.com/archives/GJQ49R20J/p1579195175088300

mpereira:
The test is currently failing because CI runs the integration tests in a Docker container with a read-only volume:
```
[21:45:29][Step 5/7] time="2020-01-21T20:45:29Z" level=error msg="Failed to write updated '../../operator/operator.yaml': open ../../operator/operator.yaml: read-only file system"
[21:45:29][Step 5/7] time="2020-01-21T20:45:29Z" level=error msg="Error overriding operatorVersion from '' to '99.99.99-testing': open ../../operator/operator.yaml: read-only file system"
```
I'll look into it.

mpereira:
I think https://github.com/mesosphere/kudo-cassandra-operator/pull/43/commits/3889befe5238a8eb0f29d3d69ef3add647d206a7 will fix it.

mpereira:
It probably did, but the currently available KUDO C* in the registry is still broken with KUDO 0.10.0. I'll fix this today.

mpereira:
The KUDO C* in the registry was fixed. Let's see how it goes now...

mpereira:
The upgrade test now passes (🎉) but there seems to be an issue with the tests which perform configuration updates now:
```
Summarizing 3 Failures:

[Fail] sanity-test [It] Updates the instance's parameters 
/kudo-cassandra-operator/tests/suites/sanity_test.go:121

[Fail] sanity-test [It] Configures Cassandra properties through custom properties 
/kudo-cassandra-operator/tests/suites/sanity_test.go:151

[Fail] sanity-test [It] Configures Cassandra JVM options through custom options 
/kudo-cassandra-operator/tests/suites/sanity_test.go:182

Ran 7 of 7 Specs in 507.276 seconds
FAIL! -- 4 Passed | 3 Failed | 0 Pending | 0 Skipped
```
I'm not sure what's causing that and will investigate tomorrow.

tests/utils/kudo/kudo.go
```diff
@@ -21,8 +24,11 @@ import (
 )
 
 var (
-	kubectlOptions *kubectl.KubectlOptions
-	kudo           *versioned.Clientset
+	kubectlOptions       *kubectl.KubectlOptions
+	kudo                 *versioned.Clientset
+	operatorYamlFilePath = "../../operator/operator.yaml"
+	// https://regex101.com/r/Ly7O1x/3/
+	semVerRegexp = `(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`
```
nfnt: Better: https://github.com/Masterminds/semver

nfnt: Oh, you don't want to check if a version is semver but instead extract the operator version from `operator.yaml`. Probably don't need this regex, you could simply assume that the string following `operatorVersion:` is semver because you can control this in this test env.

mpereira:
It's a KUDO regression: https://github.com/kudobuilder/kudo/issues/1302

mpereira:
I am planning to merge this as-is given that the test failures are caused by a KUDO regression which still needs to be fixed and released.

PR #44, opened 2020-01-24
Make operators PR remove removed files.

mpereira:
Right now the PRs are only additive. If files are removed in the kudo-cassandra-operator repository, the PRs to kudobuilder/operators aren't removing those files. This fixes that.


mpereira:
Same. Thanks for the review!

PR #46, opened 2020-01-27
Add Params for external service

ANeumann82:


tests/utils/kudo/kudo.go
```diff
@@ -293,10 +294,12 @@ func UpdateInstanceParameters(
 		return err
 	}
 
-	err = WaitForOperatorDeployInProgress(namespaceName, instanceName)
-	if err != nil {
-		log.Errorf("Error waiting for operator deploy to be in-progress: %s", err)
-		return err
+	if waitForDeploy {
```
mpereira: The "wait for deploy complete" should also be inside the conditional block, no?

ANeumann82: Mmmmm, maybe. It only assures that the plan status is in 'COMPLETE', which is ok.

Maybe the parameter should be named `waitForDeployInProgress`, because that's what I tried to accomplish.

The problem with the "WaitForOperatorDeployInProgress" is that for an update to a ConfigMap, the "IN_PROGRESS" step is so short that the test misses it.

mpereira: The idea of waiting first for the deploy to be "in-progress" and then waiting for "completed" is to follow the deploy state machine. A parameter update will trigger a deploy, which will cause the deploy status to go from "complete" -> "in-progress" -> "complete".

So the a parameter named "waitForDeploy" waiting for "in-progress" and then "complete" makes sense, because that's what one would want from a function that updates an instances parameters: change the parameter and wait for the parameter change to be reflected fully, i.e., for the deploy plan to complete.

> The problem with the "WaitForOperatorDeployInProgress" is that for an update to a ConfigMap, the "IN_PROGRESS" step is so short that the test misses it.

Is this reproducible? Where did you see it happening?

ANeumann82: Yes, generally I would agree with the COMPLETE->IN_PROGRESS->COMPLETE flow.

I've noticed the problem in this step:
https://github.com/mesosphere/kudo-cassandra-operator/pull/46/files#diff-013e06c2e358068fd6cff2d5724007a7R31

In this update, only an additional service is deployed, without updating the Pods of the StatefulSet. I assume that the Apply of the Service is quick enough that the window for "IN_PROGRESS" is very short and not reliable enough to wait for it

tests/suites/cassandra_externalservice/external_service_test.go
```diff
@@ -0,0 +1,132 @@
+package cassandra_externalservice
```
mpereira: Does it make sense to put this in `tests/suites/external_service/external_service_test.go` to maintain consistency? the "cassandra_" prefix feels extraneous...

mpereira:
```suggestion
package external_service
```

ANeumann82:
This is mostly complete - It's still missing the deletion of the external service if the feature is disabled, I'll wait for the Toggle Feature support in KUDO.

tests/suites/external_service/external_service_test.go
```diff
@@ -0,0 +1,138 @@
...
+			"EXTERNAL_NATIVE_TRANSPORT": "true",
+		})
+		Expect(err).To(BeNil())
+
+		assertNumberOfCassandraNodes(NodeCount)
+
+		log.Infof("Verify that external service is started and has 1 open port")
+		svc, err := kubernetes.GetService(Client, fmt.Sprintf("%s-svc-external", TestInstance), TestNamespace)
+		Expect(err).To(BeNil())
+		Expect(len(svc.Spec.Ports)).To(Equal(1))
```
nfnt: Can you also test that the port matches the one configured? `svc.Spec.Port[0].Name == "native-transport" && svc.Spec.Port[0].Port == EXTERNAL_NATIVE_TRANSPORT_PORT`

tests/suites/external_service/external_service_test.go
```diff
@@ -0,0 +1,138 @@
...
+			"EXTERNAL_RPC": "true",
+		})
+		Expect(err).To(BeNil())
+
+		assertNumberOfCassandraNodes(NodeCount)
+
+		log.Infof("Verify that external service is started and has 2 open ports")
+		svc, err := kubernetes.GetService(Client, fmt.Sprintf("%s-svc-external", TestInstance), TestNamespace)
+		Expect(err).To(BeNil())
+		Expect(len(svc.Spec.Ports)).To(Equal(2))
```
nfnt: Same as above, would be great to check that these ports are in fact the ones that have been configured with name and external port.

tests/suites/external_service/external_service_test.go
ANeumann82: Good point.

PR #48, opened 2020-01-29
Install KUDO inside run-tests.sh and make DS_KUDO_VERSION optional.

ANeumann82:
If DS_KUDO_VERSION is set, it's used to install KUDO on the cluster and is used for the CLI.
If its not set, use the required KUDO version for the operator


run-tests.sh
```diff
@@ -14,15 +14,32 @@ readonly kubeconfig="${KUBECONFIG:-${HOME}/.kube/config}"
...
        -e "KUBECTL_PATH=${container_vendor_directory}/kubectl.sh"  \
-       -e "DS_KUDO_VERSION=v${KUDO_VERSION}" \
+       -e "DS_KUDO_VERSION=v${DS_KUDO_VERSION}" \
+       -e "TOOLING_DIRECTORY=${container_tooling_directory}" \
+       -v "${kubeconfig}:${container_kubeconfig}:ro" \
+       -v "${project_directory}:${container_project_directory}" \
+       -v "${TOOLING_DIRECTORY}:${container_tooling_directory}" \
+       -w "${container_project_directory}" \
+       "${INTEGRATION_TESTS_DOCKER_IMAGE}" \
+       bash -c "${container_project_directory}/shared/deploy-kudo-controller-and-crds.sh"
```
mpereira: Is the idea that the command running `deploy-kudo-controller-and-crds.sh` (and the command itself) will be moved to the CI build step (where it is currently being called) instead of being in `run-tests.sh`? cc @zmalik

ANeumann82: Yes, the idea is that TC does not explicitly call deploy-kudo-controller-and-crds, but relies on run-tests.sh to do the correct setup. 

With this, it's possible to either let TC decide the KUDO version (export DS_KUDO_VERSION with a TC parameter) or let the Branch decide (no DS_KUDO_VERSION defined by TC, run-tests.sh uses the KUDO_VERSION from metadata.sh)

mpereira: Should we keep `run-tests.sh` only running tests though? This would change the development workflow as well, which also uses the script and assumes KUDO is already installed.

ANeumann82: So you'd rather have a `deploy-kudo.sh` next to `run-tests.sh`? 
I'm fine with that, that was my first idea as well, the downside is that we'd have another script in the project root. Which is admittedly is not too bad. I'll change it.

ANeumann82: It does duplicate a bit of setup work in the script, but it's ok.

ANeumann82:
I'd be happy with merging to master like this and then adjusting the TC setup.

run-tests.sh
zmalik: we decided to have separate steps also to easily identify issues
`cluster creation step` `kudo step` and actual tests
so if cluster creation step is failing we know its not a kudo or tests issue and vice versa

ANeumann82:
@zmalik It is in data-services-kudo, that's basically the `deploy-kudo-controller-and-crds.sh` from that project.

The install-kudo.sh is:
- Importing the `metadata.sh` which for C* defines the required KUDO_VERSION
- Defines a couple of paths that are required for the `deploy-kudo-controller-and-crds.sh`
- Calls `deploy-kudo-controller-and-crds.sh`

I don't think there's much I could move to the common shared directory that's not C* related.
If we were using the same setup with `metadata.sh` in all operators, it might be a good idea to move something into the shared directory, but the entry-point that includes should still be in the project itself, not in the shared-folder.

zmalik:
what is Cassandra related here that doesn't apply for Kafka or Spark for example? can you please explain?

ANeumann82:
Well... Compared to the `run-tests.sh`, the only thing that is *not* Cassandra-Related are these lines:

```
export DS_KUDO_VERSION="${DS_KUDO_VERSION:-${KUDO_VERSION}}"

if [[ ${DS_KUDO_VERSION:0:1} == "v" ]] ; then DS_KUDO_VERSION="${DS_KUDO_VERSION:1}"; fi
```

Everything else you can find as well in the `run-tests.sh`, which is specific to the C*-Repository.

I don't think we could use the `install-kudo.sh` in the kafka-repository at the moment, as the kafka-repository doesn't know the minimum KUDO version required to run the operator, at least not in an exported ENV-Variable.

zmalik:
I'm just talking about installing KUDO and I still don't see why installing KUDO would need Cassandra specific code

> Everything else you can find as well in the run-tests.sh, which is specific to the C*-Repository.

`run-tests.sh` is assuming that KUDO is already installed. 
If you are moving Cassandra specific things to another script, please don't call it `install-kudo.sh`as it can be really confusing

tests/utils/kudo/kudo.go
```diff
@@ -232,7 +232,7 @@ func WaitForOperatorDeployComplete(
 	namespaceName string, instanceName string,
 ) error {
 	// 5 minutes.
```
porridge: Please either update or remove this comment.

tests/utils/kudo/kudo.go
```diff
@@ -215,7 +215,7 @@ func WaitForOperatorDeployInProgress(
 	namespaceName string, instanceName string,
 ) error {
 	// 30 seconds.
```
porridge: Please either update or remove this comment.

tests/suites/sanity_test.go
```diff
@@ -190,6 +175,21 @@ var _ = Describe(TestName, func() {
 		assertNumberOfCassandraNodes(NodeCount)
 	})
 
+	It("Scales the instance's number of nodes", func() {
```
porridge: What is the reason for moving this part?

install-kudo.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Note: DS_KUDO_VERSION is used by the shared data-services-kudo tooling.
+# DS_KUDO_VERSION *may* be set by TeamCity Jobs if a fixed KUDO version is preferred for the test execution
+# If not DS_KUDO_VERSION is set, we use and install the required KUDO version from the operator
+export DS_KUDO_VERSION="${DS_KUDO_VERSION:-${KUDO_VERSION}}"
+
+# Strip 'v' from "v0.10.1" if it's there.
+if [[ ${DS_KUDO_VERSION:0:1} == "v" ]] ; then DS_KUDO_VERSION="${DS_KUDO_VERSION:1}"; fi
+
+docker run \
```
porridge: This docker invocation, and in fact most of this file seems nearly identical to the one in `run-tests.sh`.
Please refactor this so the code is not duplicated.

ANeumann82:
Not required anymore, superseded by #52

PR #50, opened 2020-01-30
Add PDB and UpdateStrategy to Parallel

ANeumann82:
UpdateStrategy Parallel should make setup of the operator much faster.

Add PodDisruptionBudget


ANeumann82:
This is going to be included in the backup PR

PR #53, opened 2020-02-10
Add test to change mem and CPU for nodes

ANeumann82:


tests/suites/sanity/sanity_test.go
```diff
@@ -90,6 +93,42 @@ var _ = Describe(TestName, func() {
...
+		pod, err := kubernetes.GetPod(Client, TestInstance+"-node-0", TestNamespace)
+
+		Expect(err).To(BeNil())
+		Expect(pod).To(Not(BeNil()))
+
+		Expect(pod.Spec.Containers[0].Resources.Requests.Cpu().AsDec().UnscaledBig()).To(Equal(big.NewInt(int64(newCpu))))
+		Expect(pod.Spec.Containers[0].Resources.Requests.Memory().AsDec().UnscaledBig()).To(Equal(big.NewInt(int64(newMemBytes))))
+
+		Expect(pod.Spec.Containers[0].Resources.Limits.Cpu().AsDec().UnscaledBig()).To(Equal(big.NewInt(int64(newCpuLimit))))
+		Expect(pod.Spec.Containers[0].Resources.Limits.Memory().AsDec().UnscaledBig()).To(Equal(big.NewInt(int64(newMemBytes))))
```
nfnt:
```suggestion
		Expect(pod.Spec.Containers[0].Resources.Limits.Memory().AsDec().UnscaledBig()).To(Equal(big.NewInt(int64(newMemLimitMiB))))
```

ANeumann82: That was actually intentional, as we're setting 3192, but the returned value from k8s is (3192 * 1024 * 1024).

I've changed it now so that this is correct and more explicit.

PR #54, opened 2020-02-12
Add prometheus integration test to sanity

ANeumann82:

### DIFF
--- a/tests/curl/curl.go
+++ b/tests/curl/curl.go
@@ -2,13 +2,14 @@ package curl
 
 import (
 	"fmt"
+	"strings"
+
 	"github.com/kudobuilder/test-tools/pkg/client"
 	"github.com/kudobuilder/test-tools/pkg/cmd"
 	"github.com/kudobuilder/test-tools/pkg/kubernetes"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/uuid"
-	"strings"
 )
 
 func RunCommand(client client.Client, namespace string, arguments ...string) (string, string, error) {
--- a/tests/curl/curl.go
+++ b/tests/curl/curl.go
@@ -1,0 +1,64 @@
+package curl
+
+import (
+	"fmt"
+	"github.com/kudobuilder/test-tools/pkg/client"
+	"github.com/kudobuilder/test-tools/pkg/cmd"
+	"github.com/kudobuilder/test-tools/pkg/kubernetes"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"strings"
+)
+
+func RunCommand(client client.Client, namespace string, arguments ...string) (string, string, error) {
+	podTpl := v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			// TODO: When this is more generic, pod name should probably be randomized
+			Name:      "curl-pod",
+			Namespace: namespace,
+		},
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{
+				{
+					Name:    "curl",
+					Image:   "curlimages/curl",
+					Command: []string{"sleep", "300"},
+				},
+			},
+		},
+	}
+
+	pod, err := kubernetes.NewPod(client, podTpl)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to create new curl pod: %v", err)
+	}
+	defer func() {
+		_ = pod.Delete()
+	}()
+
+	for {
+		if pod.Status.Phase == v1.PodRunning {
+			break
+		} else {
+			err = pod.Update()
+			if err != nil {
+				return "", "", fmt.Errorf("failed to update pod: %v", err)
+			}
+		}
+	}
+
+	var stdout strings.Builder
+	var stderr strings.Builder
+
+	cmd := cmd.New("curl").
+		WithArguments(arguments...).
+		WithStdout(&stdout).
+		WithStderr(&stderr)
+
+	err = pod.ContainerExec("curl", cmd)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to exec in container: %v", err)
+	}
+
+	return stdout.String(), stderr.String(), nil
+}
--- a/tests/curl/curl.go
+++ b/tests/curl/curl.go
@@ -7,14 +7,16 @@ import (
 	"github.com/kudobuilder/test-tools/pkg/kubernetes"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/uuid"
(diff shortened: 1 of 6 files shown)


tests/suites/sanity/sanity_test.go
```diff
@@ -29,8 +31,9 @@ var (
 	TestInstance        = fmt.Sprintf("%s-instance", OperatorName)
 	KubeConfigPath      = os.Getenv("KUBECONFIG")
 	OperatorDirectory   = os.Getenv("OPERATOR_DIRECTORY")
-	// TODO(mpereira): read NodeCount from params.yaml.
-	NodeCount = 3
+
+	// Node Count of 1 for Sanity test to have the tests a little bit faster
+	NodeCount = 1
```
mpereira: With a node count of 1, the value provided by `assertNumberOfCassandraNodes` (which effectively gets the output from `nodetool status`) is greatly diminished, because now it's not testing that the nodes are forming a cluster at all.

tests/suites/sanity/sanity_test.go
```diff
@@ -90,6 +93,55 @@ var _ = Describe(TestName, func() {
...
+				pod, err = kubernetes.GetPod(Client, pod.Name, pod.Namespace)
+				Expect(err).To(BeNil())
+			}
+		}
+
+		var stdout strings.Builder
+		var stderr strings.Builder
+
+		cmd := cmd.New("curl").
+			WithArguments("-s", "prometheus-kubeaddons-prom-prometheus.kubeaddons.svc.cluster.local:9090/api/v1/query?query=cassandra").
```
mpereira: Very nice! Does it also make sense to create a new "module", like `prometheus.go` and encapsulate the usage of the `curl` container? Or at least to extract the Prometheus URL to somewhere else, maybe a constant or function?

tests/suites/sanity/sanity_test.go
mpereira: This is a very important and valuable test to maintain, IMO, as changes to KUDO or the KUDO C* operator itself might cause the cluster to not form anymore. This already happened before and the tests caught it.

tests/suites/sanity/sanity_test.go
ANeumann82: Yeah, that would probably make sense to abstract into the test-tools

tests/suites/sanity/sanity_test.go
ANeumann82: Yes, I do agree that it is important to check that the cluster is actually formed, but I think it should be enough to do this in the very last step of the sanity-test. Otherwise every Parameter-Update is triple the time as the rolling update to the cluster restarts every node.

mpereira: One could also argue that it's important to check the changes to the cluster like parameter updates happen to a formed cluster, not necessarily an isolated node. It feels to me that the benefit of having a faster integration test suite doesn't outweigh the downside of making the test checks less valuable.

The existing functions that only query 0th pods could also be improved, thinking about it...

tests/suites/sanity/sanity_test.go
```diff
@@ -90,6 +96,55 @@ var _ = Describe(TestName, func() {
...
+		}
+
+		pod, err := kubernetes.NewPod(Client, podTpl)
+		Expect(err).To(BeNil())
+
+		for {
+			if pod.Status.Phase == v1.PodRunning {
+				break
+			} else {
+				pod, err = kubernetes.GetPod(Client, pod.Name, pod.Namespace)
```
nfnt: You could also run
```suggestion
				err = pod.Update()
```

tests/suites/sanity/sanity_test.go
```diff
@@ -90,6 +96,55 @@ var _ = Describe(TestName, func() {
...
+						Command: []string{"sleep", "300"},
+					},
+				},
+			},
+		}
+
+		pod, err := kubernetes.NewPod(Client, podTpl)
+		Expect(err).To(BeNil())
+
+		for {
```
nfnt: IMO, this is okay for this test. But if for any reason the pod doesn't get into `PodRunning` phase, we would loop until we run into a timeout. I'd rather have a loop that can timeout here, like in https://github.com/kudobuilder/test-tools/blob/master/pkg/kudo/instance.go#L65. Though this looks like a feature for `test-tools`, you could also create an issue there ;)

tests/suites/sanity/sanity_test.go
nfnt: +1 for prometheus helpers in `test-tools`!

ANeumann82: Well, rather curl-helpers. Maybe Prometheus as well...

tests/curl/curl.go
```diff
@@ -0,0 +1,67 @@
...
+			},
+		},
+	}
+
+	pod, err := kubernetes.NewPod(client, podTpl)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to create new curl pod: %v", err)
+	}
+	defer func() {
+		_ = pod.Delete()
```
nfnt: Nit: Let's at least log something if the pod deletion fails. Or use named result value as a workaround to capture the error like in this example: https://play.golang.org/p/PZGn5-1TXv4.

tests/curl/curl.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/kudobuilder/test-tools/pkg/client"
+	"github.com/kudobuilder/test-tools/pkg/cmd"
+	"github.com/kudobuilder/test-tools/pkg/kubernetes"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/uuid"
+)
+
+func RunCommand(client client.Client, namespace string, arguments ...string) (string, string, error) {
```
nfnt: Nit: Let's just call this `Run`

tests/prometheus/prometheus.go
```diff
@@ -0,0 +1,45 @@
...
+	Type   string                   `json:"resultType"`
+	Result []PrometheusSearchResult `json:"result"`
+}
+
+type PrometheusSearchResult struct {
+	Metric map[string]interface{} `json:"metric"`
+	Value  []interface{}          `json:"value"`
+}
+
+func QueryForStats(client client.Client, namespace string, baseURL, query string) (*PrometheusSearch, error) {
```
nfnt: Does the namespace matter here? It's [already part of the DNS](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#namespaces-and-dns), or, in case of a FQDN, namespaces shouldn't matter.

tests/curl/curl.go
ANeumann82: Fancy. Didn't know you could do that. I added a log though, as it's a warning at most and nothing critical.

tests/prometheus/prometheus.go
ANeumann82: It doesn't, really. It was only required for the Curl-Runner, to know in which namespace to launch the temporary pod. I've refactored that so it's not part of the QueryForStats anymore

PR #55, opened 2020-02-17
Configure datacenter and rack when using 'GossipingPropertyFileSnitch' 

nfnt:
The 'GossipingPropertyFileSnitch' propagates rack and datacenter information in a Cassandra cluster. This topology information is used by Cassandra's replication strategy.

Note for reviewers:
This uses `ginkgo` in a different way than the other tests. I made the tests more BDD like and used tighter scoping of variables. Let me know what you think and if these patterns should be used in the other test suites as well.


nfnt:
CI is currently failing to push Docker images due to https://www.mail-archive.com/user@cassandra.apache.org/msg60484.html. tl;dr: Upstream docker images need new GPG keys because of changed redirects of the Cassandra debian repo.

operator/params.yaml
```diff
@@ -60,6 +60,14 @@ parameters:
     description: "When a pod starts and the probe fails, `failure_threshold` attempts will be made before marking the pod as 'unready'."
     default: "3"
 
+  - name: NODE_DATACENTER
```
nfnt: We might not want the user to set these parameters. I'm currently figuring out the [cluster's topology](https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#topologykubernetesiozone) should and can be used instead.

nfnt: Yeah, this is a bit too simple. We want to support setups where, e.g., 2 nodes are in rack 1 and another node is in rack 2. This demands a more sophisticated configuration and would result in multiple `StatefulSets`, one for each datacenter/rack.

nfnt:
Added the `wip` label again because more work is needed to support other scenarios.

nfnt:
Added my latest WIP state. Currently testing locally, the CI hasn't been updated yet. The default parameters have been changed to reflect the test setup, because inputting YAML parameter values during `kubectl kudo install` isn't well supported by KUDO right now. The deployment is currently failing because the seed list hasn't been updated yet.

operator/templates/stateful-set.yaml
```diff
@@ -196,50 +256,90 @@ spec:
...
+              cpu: "{{ $.Params.PROMETHEUS_EXPORTER_CPU_LIMIT_MC }}m"
           volumeMounts:
             - name: cassandra-exporter-config-yml
               mountPath: /etc/cassandra_exporter/config.yml
               subPath: config.yml
         {{ end }}
+      {{ if $.Params.NODE_TOPOLOGY }}
+      initContainers:
+        - name: node-resolver
+          image: bitnami/kubectl:latest
```
nfnt: Using `latest` could be problematic, as this might differ from the version of the cluster. In the worst case this could cause issues. We should either use a tag that matches the min-required version of Kubernetes as stated in `operator.yaml` or make the tag a parameter.

ANeumann82: Good point. I'm going to change that

PR #58, opened 2020-02-21
Allow JMX rpc to be accessed from within the cluster

ANeumann82:
This is prework for backup/restore, so we can control the cluster/node from outside the running pod.


operator/params.yaml
```diff
@@ -143,6 +143,10 @@ parameters:
     description: "The JMX port that will be used to interface with the Cassandra application."
     default: "7199"
 
+  - name: JMX_LOCAL_ONLY
+    description: "If true, the JMX port will only be opened on localhost and not be available to the cluster"
+    default: "false"
```
nfnt: Let's use the more restrictive local-only configuration as default.
```suggestion
    default: "true"
```

operator/templates/cassandra-env-sh.yaml
```diff
@@ -242,31 +242,29 @@ data:
...
-    else
+    {{ else }}
       JVM_OPTS="$JVM_OPTS -Dcassandra.jmx.remote.port=$JMX_PORT"
       # if ssl is enabled the same port cannot be used for both jmx and rmi so either
       # pick another value for this property or comment out to use a random port (though see CASSANDRA-7087 for origins)
       JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.rmi.port=$JMX_PORT"
 
       # turn on JMX authentication. See below for further options
-      JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.authenticate=true"
+      JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.authenticate=false"
```
nfnt: Disabling authentication and SSL seems very insecure. This should be configurable and enabled by default.

operator/params.yaml
ANeumann82: Yeah, I thought about that this morning as well. This will be a prerequisite for enabling backup & restore though, it might be a good idea to have some feature in KUDO that allows checking of depending parameters, i.e. if you have `ENABLE_BACKUP_RESTORE=true` you also need to have `JMX_LOCAL_ONLY=false`.

Changed it to default false anyway.

operator/templates/cassandra-env-sh.yaml
ANeumann82: There is an explicit ticket for this. I'd prefer to focus on the actual backup & restore for now, as this might lead to bigger changes in the operator itself. 
I do agree that this should be configurable and be enabled by default though.

PR #60, opened 2020-02-28
Backup with Medusa to S3

ANeumann82:
SSL support is prepared, waiting for https://github.com/thelastpickle/cassandra-medusa/pull/119 to be merged and released


images/Dockerfile.medusa-backup-fromsrc
```diff
@@ -0,0 +1,18 @@
+FROM cassandra:3.11.5
+
+COPY medusa/cassandra-sudoer /etc/sudoers.d/
+
+COPY medusa/medusa-src /medusa-src
```
ANeumann82, line 5: This Dockerfile can probably be removed at some point. It's not included in the build.sh, but very useful to test new medusa versions from source.

operator/templates/backup-job.yaml
```diff
@@ -0,0 +1,22 @@
+{{- $root := . -}}
```
nfnt, line 1: Nit: Not needed, the root is always accessibly with `$`. E.g. ``$.Name`. See `$.Params.KUBECTL_VERSION` used in this file.

operator/templates/node-scripts.yaml
```diff
@@ -33,3 +33,48 @@ data:
     dc=$CASSANDRA_DATACENTER
     rack=$RACK
     EOF
+  node-token-save.sh: |
+    # Used to capture the token map from a newly created cluster and save it
+    # Not used at the moment
+    if [ ! -f /var/lib/cassandra/token_map ]; then
+        while [ -z $NODE_TOKENS ]; do
+            NODE_ID=`nodetool info | grep ID | sed -n -e 's/ID[[:space:]]*\:[[:space:]]*\(.*\)$/\1/p'`
+            NODE_IP=`nodetool status | grep $NODE_ID | sed -n -e 's/UN[[:space:]]*\([0-9.]*\)[[:space:]]*.*/\1/p'`
```
nfnt, line 42: Nit: Do we have to run `nodetool info` and `nodetool status` for every iteration of the while-loop? How about running them once and only run `nodetool ring` in the loop? I'm not sure how many iterations such a loop has, but `nodetool` is quite expensive and this could reduce performance if there are many iterations.

ANeumann82: tbh, I'm not sure anymore. I think I've used that in one iteration of development to save the token_map. The while loop is there so the script waits until the cluster is fully up and running, but... Yeah, that could probably be optimized. 

The script is currently not used, i just didn't want to delete it yet, as figuring out the sed regexps for the fields was a PITA. 
I'll probably delete the script once the branch is merged - then it's still available in the git history.

operator/templates/backup-job-rbac.yaml
```diff
@@ -0,0 +1,33 @@
+{{ if eq .Params.BACKUP_RESTORE_ENABLED "true" }}
```
zmalik, line 1: can this be a toggle task? when we disable `BACKUP_RESTORE_ENABLED ` using `false` these resources won't be cleaned up

operator/templates/medusa-config-ini.yaml
```diff
@@ -0,0 +1,101 @@
+{{ if eq .Params.BACKUP_RESTORE_ENABLED "true" }}
```
zmalik, line 1: same here for the toggle task

tests/suites/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,114 @@
...
+	_ = kubernetes.CreateNamespace(Client, TestNamespace)
+})
+
+var _ = AfterSuite(func() {
+	_ = Operator.Uninstall()
+	_ = kubernetes.DeleteNamespace(Client, TestNamespace)
+})
+
+// This test is disabled for now, as the operator does not really support upgrading yet
+//func TestService(t *testing.T) {
```
zmalik, line 44: we just moved this to a new file?

ANeumann82: Yes

PR #61, opened 2020-02-28
Make sure that the "compiled" files match their templates.

porridge:
To keep CI green:
- switch from using `golang:1.13.1-stretch` test image to a customized one with `gettext-base` installed so we can actually run `envsubst`
- update templates which went out of sync

### DIFF
--- a/images/Dockerfile.integration-tests
+++ b/images/Dockerfile.integration-tests
@@ -1,0 +1,6 @@
+FROM golang:1.13.1-stretch
+
+# For envsubst:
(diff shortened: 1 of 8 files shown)


ANeumann82:
Oh that is a good check!

porridge:
```
[13:58:44][Step 5/7] ./tools/compile_templates.sh: line 59: envsubst: command not found
```
awesome :-(

porridge:
OK, this now actually works, please review @nfnt @ANeumann82

PR #63, opened 2020-03-11
Add run-pre-checks.sh

ANeumann82:
- Template check
- goimports check

### DIFF
--- a/run-pre-checks.sh
+++ b/run-pre-checks.sh
@@ -8,16 +8,15 @@ readonly project_directory="$(readlink -f "${script_directory}")"
 # shellcheck source=metadata.sh
 source "${project_directory}/metadata.sh"
 
+FAILED=0
+
 go get golang.org/x/tools/cmd/goimports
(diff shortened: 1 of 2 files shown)


run-pre-checks.sh
```diff
@@ -0,0 +1,18 @@
...
+readonly project_directory="$(readlink -f "${script_directory}")"
+
+# shellcheck source=metadata.sh
+source "${project_directory}/metadata.sh"
+
+FAILED=0
+
+cd "${project_directory}"
+./tools/compile_templates.sh --check-only
+FAILED=${FAILED-$?}
```
porridge: This is not even reachable on failure with `set -e`. Please delete.

ANeumann82: tbh, I'd rather remove the "-e" and handle exit codes manually. This would allow us to have multiple checks that are all executed, regardless if one of them fails...

porridge: That can still be done with an `if`.

PR #65, opened 2020-03-12
Add Flag for local testing with limited resources

ANeumann82:
Also cleaned up some tests by using "By" instead of "It"

### DIFF
--- a/tests/suites/sanity/sanity_test.go
+++ b/tests/suites/sanity/sanity_test.go
@@ -10,35 +10,51 @@ import (
 	"testing"
 	"time"
 
-	"github.com/mesosphere/kudo-cassandra-operator/tests/curl"
+	"github.com/onsi/ginkgo/reporters"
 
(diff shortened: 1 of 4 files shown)


tests/suites/sanity/sanity_test.go
```diff
@@ -65,53 +83,23 @@ var _ = Describe(TestName, func() {
 
 		err = Operator.Instance.WaitForPlanComplete("deploy")
 		Expect(err).To(BeNil())
-
 		assertNumberOfCassandraNodes(NodeCount)
-	})
 
-	It("Upgrades the running operator instance from a directory", func() {
-		before, _, err := cassandra.OverrideOperatorVersion(TestOperatorVersion)
```
zmalik: you can remove the function `OverrideOperatorVersion` as after this removal it shouldn't be used anywhere else.

ANeumann82: Good point. I want to keep it for now as the upgrade test is coming back, as an extra test though

PR #66, opened 2020-03-18
Use `$IMAGE_DISAMBIGUATION_SUFFIX` if set.

porridge:
If the `$IMAGE_DISAMBIGUATION_SUFFIX` environment variable is explicitly set (normally only by TeamCity), then we:
- compile templates before testing and building images,
- use the variable in `metadata.sh` to prevent clashes of docker images between multiple in-flight PRs.

We temporarily unset the variable when checking template expansion in `run-pre-checks.sh`.

Fixes: https://github.com/kudobuilder/operators/issues/234
TODO:
- [x] get the `shared` part of this logic (which actually exports the variable) merged: https://github.com/mesosphere/data-services-kudo/pull/56
- [x] rebase on it


porridge:
It seems GitHub is having problems assigning reviewers to PRs.
@nfnt @zmalik please review.

metadata.sh
```diff
@@ -68,7 +68,7 @@ export PROMETHEUS_EXPORTER_DOCKER_IMAGE="${PROMETHEUS_EXPORTER_DOCKER_IMAGE_NAME
 export INTEGRATION_TESTS_DOCKER_IMAGE_FROM="golang:1.13.1-stretch"
 export INTEGRATION_TESTS_DOCKER_IMAGE_NAMESPACE="mesosphere"
 export INTEGRATION_TESTS_DOCKER_IMAGE_NAME="kudo-cassandra-tests"
-export INTEGRATION_TESTS_DOCKER_IMAGE_TAG="latest"
+export INTEGRATION_TESTS_DOCKER_IMAGE_TAG="latest${IMAGE_DISAMBIGUATION_SUFFIX:-}"
```
nfnt: Nit: Let's remove `latest` here completely and use a different that than `latest` if `IMAGE_DISAMBIGUATION_SUFFIX` isn't set.

porridge: TBH I'm not even sure why we're pushing this image at all :thinking:

PR #69, opened 2020-03-30
Bump to KUDO 0.11.1

porridge:
- also update the stale URL in a way which will not go stale ever


porridge:
Investigating these failures via https://github.com/mesosphere/kudo-cassandra-operator/pull/70

kensipe:
superseded by #78

PR #73, opened 2020-04-02
Configure liveness probes for pods

nfnt:
Liveness probes check that the Cassandra node running in the pod is responsive by calling 'nodetool info'. It does not consider the node's status provided by 'nodetool status' for liveness. As such it's only concerned by the liveness of the pods application.

### DIFF
--- a/operator/params.yaml
+++ b/operator/params.yaml
@@ -53,11 +53,31 @@ parameters:
     default: "60"
 
   - name: NODE_READINESS_PROBE_SUCCESS_THRESHOLD
-    description: "Minimum consecutive successes for the probe to be considered successful after having failed."
+    description: "Minimum consecutive successes for the readiness probe to be considered successful after having failed."
     default: "1"
 
   - name: NODE_READINESS_PROBE_FAILURE_THRESHOLD
-    description: "When a pod starts and the probe fails, `failure_threshold` attempts will be made before marking the pod as 'unready'."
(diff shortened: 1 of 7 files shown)


operator/params.yaml
```diff
@@ -53,11 +53,31 @@ parameters:
...
   - name: NODE_READINESS_PROBE_FAILURE_THRESHOLD
-    description: "When a pod starts and the probe fails, `failure_threshold` attempts will be made before marking the pod as 'unready'."
+    description: "When a pod starts and the readiness probe fails, `failure_threshold` attempts will be made before marking the pod as 'unready'."
+    default: "3"
+
+  - name: NODE_LIVENESS_PROBE_INITIAL_DELAY_S
+    description: "Number of seconds after the container has started before the liveness probe is initiated."
+    default: "15"
+
+  - name: NODE_LIVENESS_PROBE_PERIOD_S
```
zmalik, line 67: nodetool checks are quite heavy. 
Moreover with liveness check, if it starts happening every 20 seconds. We should open an issue to implement this without calling nodetool. Using any jmx agent like jolokia should be enough.

nfnt: Then maybe a [TCP liveness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-tcp-liveness-probe) would be enough. This would check for the availability of the JMX.

nfnt: I had to revert this approach. Trying to connect via TCP to the JMX port is being refused. I'll open an issue for us to look at JMX agent like you suggested above.

PR #78, opened 2020-04-07
Bump to kudo v0.11.1

porridge:
This PR depends on a couple of pending changes in dependencies:
- [x] wait for `UninstallWaitForDeletion` addition to be merged in `test-tool`. https://github.com/kudobuilder/test-tools/pull/25
- [x] wait for https://github.com/mesosphere/data-services-kudo/pull/58 to be merged
- [x] update this PR to point at the above merged changes, rather than feature branches

Unfortunately this is more than a simple bump, and required quite a bit of debugging to make it work.

TL;DR: until `v0.11.0` KUDO controller manager was quite slow, but in `v0.11.1` it is significantly faster, causing it always to win a race which turns a [deficiency in stateful set controller](https://github.com/kubernetes/kubernetes/issues/74374) into a deadlock.

# An amazing tale of distributed system adventures

## What the KUDO-cassandra TLS test suite does

The relevant piece is that [it](https://github.com/mesosphere/kudo-cassandra-operator/blob/master/tests/suites/tls/tls_test.go) performs the following sequence of steps *three times*, in quick succession:
1. Install kudo-cassandra operator, operatorversion and instance
1. Wait for instance to become healthy. An important detail is that kudo-cassandra uses a stateful set, here `cassandra-instance-node`, with two pods, `cassandra-instance-node-X` each of which requires an auto-provisioned persistent volume.
1. [remove](https://github.com/mesosphere/kudo-cassandra-operator/blob/78b22ddc3ab3883e099acd95bddae59ea8d5640e/tests/cassandra/cassandra.go#L212) kudo-cassandra operator, operatorversion and instance
1. [remove](https://github.com/mesosphere/kudo-cassandra-operator/blob/78b22ddc3ab3883e099acd95bddae59ea8d5640e/tests/cassandra/cassandra.go#L216) both persistentvolumeclaims

## The failure

After bumping the kudo version it turned out that this test *always* failed with a timeout waiting for install (step 1) to finish but *only on the second and third occurrences* of this sequence. Initial inspection showed that the `cassandra-instance-node-1` pod fails to schedule due to a missing persistent volume.

## The investigation

I suspected that the reason I'm seeing the failure so reliably is that the timing changed significantly in the newer KUDO version.

I initially wasted some time barking up the [wrong tree](https://github.com/kubernetes/kubernetes/issues/76703) that is still present in k8s 1.15 that we're using. Unfortunately the `Event`s logged by various controllers were not enough to confirm or disprove the theory that this was the root cause.

It was only after a more careful and organized debugging session that I understood what is going on, and realized that the PVC protection controller was working as intended in this case. The tool which did contain all the necessary information was the `kube-apiserver` audit log (luckily already turned on by konvoy and dumped by `konvoy diagnose`). I also needed to improve my `jq` skills to be able to understand what it was recording.

First I aggreated the log from three master nodes into a single chronological one with:

```
cat */kube-apiserver-audit.log| jq -c '[.requestReceivedTimestamp,.]' | sort | jq -c '.[1]' > api-log.json
```

Next, since I knew the problem was somewhere around pods, persistentvolumes and persistentvolumeclaims, I looked at a chronological overview of mutations of these kinds with something like the following - refining the query step by step as I focused more and more closely on the interesting part:

```
cat api-log.json | jq '
  select(
    (.objectRef.resource=="pods" and .objectRef.name=="cassandra-instance-node-1")
    or
    (.objectRef.resource=="persistentvolumeclaims" and .objectRef.name=="var-lib-cassandra-cassandra-instance-node-1")
    or
    (.objectRef.resource=="persistentvolumes" and .objectRef.name=="pvc-2aaa4cff-ec2b-47ef-9740-e36778265492")
    or
    (.objectRef.resource=="statefulsets" and .objectRef.name=="cassandra-instance-node")
    or
    (.objectRef.resource=="instances" and .objectRef.name=="cassandra-instance"))' | \
jq -r '[.requestReceivedTimestamp,.user.username,.verb,.responseStatus.code,.objectRef.resource,.objectRef.name,.objectRef.subresource?]|@csv'  > report.csv
```

I used LibreOffice CSV import for viewing and annotating the output, for clear tabular format.

Other useful filters include:
- by actor: `select(.user.username=="kubernetes-admin" or .user.username=="system:serviceaccount:kudo-system:kudo-manager")`
- by time: `select(.requestReceivedTimestamp >= "2020-04-06T06:37:23" and .requestReceivedTimestamp <= "2020-04-06T06:37:51")`
- dumping individual records into files for diff viewing using meld, to understand the intents of individual mutations: `jq 'select(.requestReceivedTimestamp=="2020-04-06T06:37:23.416332Z")' api-log.json`

## The issue

I eventually tracked down the deadlock to be:

1. test DELETEs the instance and PVCs within 7ms
1. test CREATEs a new instance 20ms later
1. KUDO manager takes over the stateful set just 34ms later and PATCHes it to point at its new parent instance, with a new `kudo.dev/last-plan-execution-uid` annotation
1. statefulset controller begins to roll out this change: DELETEs the last pod (`cassandra-instance-node-1`)
1. ~30s later the kubelet get rids of the pod
1. 25ms later `pvc-protection-controller` removes its finalizer from the PVC now that the pod is gone, note that the PVC is still not gone at this point
1. 5ms later statefulset controller creates a new pod, but does not do anything the PVC, since it is still present
1. 14ms later persistent-volume-binder marks the PV as released and it is eventually GC-ed

The issue is that now that the pod is created, stateful set controller never considers even taking a look at the PVC, and the pod goes pending forever. This is described (among other things) in https://github.com/kubernetes/kubernetes/issues/74374 and there is even a [potential fix](https://github.com/kubernetes/kubernetes/issues/74374#issuecomment-506939665) waiting for almost a year already...

## The trigger

I still did not understand exactly how the KUDO version bump tickled this issue. However finding out was relatively easy now that I knew what tools to use. I triggered a run on `master` and collected the audit log. It turns out that in case of `v0.11.0` the sequence looks like this:

1. test DELETEs the instance and PVCs within 84ms
1. test CREATEs a new instance 140ms later
1. ~700ms later garbage collector DELETES the statefulset and pods altogether
1. only 44 *seconds* later does kudo-manager wake up enough to re-create the statefulset afresh


tests/cassandra/cassandra.go
```diff
@@ -209,7 +210,10 @@ func Cqlsh(client client.Client, instance kudo.Instance, cql string) (string, er
 }
 
 func Uninstall(client client.Client, operator kudo.Operator) error {
-	if err := operator.Uninstall(); err != nil {
+	// This wait is necessary to avoid tickling an issue in stateful set controller,
+	// which gets stuck with a pod but no PVC when KUDO is quick to process the instance delete
+	// (and create by subsequent test).
+	if err := operator.UninstallWaitForDeletion(5 * time.Minute); err != nil {
```
kensipe, line 216:
```suggestion
	timeout := 5 * time.Minute
	if err := operator.UninstallWaitForDeletion(timeout); err != nil {
```

PR #79, opened 2020-04-07
Linted project

kensipe:
Signed-off-by: Ken Sipe <kensipe@gmail.com>

**NOTE**: The following are details of the PR template... which are completely broken.   Both scripts below are broken on master.  It isn't clear if the scripts are needed or if they should be deleted... certainly the PR template needs to fixed.  Waiting for resolution of scripts first.

3. Please make sure that that `git status` looks clean after running
   `./tools/compile_templates.sh`. If there's a diff you might need to commit
   further changes.

4. Please make sure that that `git status` looks clean after running
   `./tools/generate_parameters_markdown.py`. If there's a diff you might need
   to commit further changes.


zmalik:
scripts should be fixed. As Cassandra repo is using templating to generate final operator content.

.gitignore
```diff
@@ -1 +1,2 @@
 *-junit.xml
+tests/.idea/
```
porridge, line 2:
```suggestion
/tests/.idea/
```

kensipe: this change would not be appropriate... it is a relative path we are looking to ignore

porridge: Actually, feel free to ignore this. Looks like for multi-component paths the semantics is the same.

I thought it differs like for simple paths, where `/foo` only ignores `foo` that is in the root of the repo, while `foo` ignores all `foo` files everywhere in the repo.

PR #84, opened 2020-04-09
Improvements to docker image.

porridge:
tools/Dockerfile:
- cleanup pubkey.gpg and break long command
- install pyyaml from a deb rather than pip, since it's available and
  likely to be more stable

tools/docker.sh:
- use errexit to catch docker build failures rather than silently use a
  stale image
- echo errors to stderr
- use a $REPO_ROOT variable rather than TOOL_DIR and drop the "cd"s to
  decrease congnitive burden


kensipe:
merging into the other pr branch!

PR #86, opened 2020-04-14
[D2IQ-60578] Use debug.CollectArtifacts

porridge:


nfnt:
Which artifacts in Teamcity are created from this? I can see `kudo.zip`, `konvoy.zip`, and `kubernetes.zip`. It doesn't look like there are artifacts created from the test suites.

porridge:
> Which artifacts in Teamcity are created from this? I can see `kudo.zip`, `konvoy.zip`, and `kubernetes.zip`. It doesn't look like there are artifacts created from the test suites.

`operator.zip`:
Zrzut ekranu z 2020-04-16 16-38-50

nfnt:
I don't see `operator.zip` in the artifacts of the latest run (# 1280): https://teamcity.mesosphere.io/viewLog.html?buildId=2704851&buildTypeId=Frameworks_DataServices_Kudo_Cassandra_Pr_CassandraPrKonvoyKudo_2&tab=artifacts

porridge:
Right, because in order to have a green run I removed the commit which made cassandra always fail, so the tests pass, so the diagnostics is not collected. :-)

nfnt:
I see. How about always collecting these artifacts? There's some value to look into these even when tests suceeded. E.g., when writing new tests that suddenly succeed in situation where one would still expect them to fail.

porridge:
I was afraid this would be more annoying than helpful, but with 4 seconds per collection maybe it's not so bad. Done, PTAL.

PR #88, opened 2020-04-16
Support password authentication

nfnt:
Calls by the operator to `nodetool` are authenticated with credentials provided as a secret by the admin.


operator/templates/repair-job.yaml
```diff
@@ -12,7 +12,21 @@ spec:
     spec:
       containers:
         - name: repair-job
-          image: bitnami/kubectl:1.18.0
-          command: [ "kubectl", "exec", "{{ $.Params.REPAIR_POD }}", "--", "nodetool", "repair" ]
+          image: bitnami/kubectl:latest
```
ANeumann82, line 15: I'm not sure using `latest` here is a good idea. We might want to make this a fixed version, otherwise we probably run into issues when kubectl 1.19, 1.20, etc is release

operator/templates/stateful-set.yaml
```diff
@@ -356,6 +364,11 @@ spec:
             name:  {{ $.Name }}-generate-tls-artifacts-sh
             defaultMode: 0755
         {{ end }}
+        {{ if $.Params.AUTHENTICATION_CREDENTIALS_SECRET }}
+        - name: {{ $.Params.AUTHENTICATION_CREDENTIALS_SECRET }}
```
ANeumann82, line 368: The name should probably not be the param, right? I'm not sure why it's used that way for the TLS_SECRET above, but the name here is just a fixed string that is used to reference the secret in the volumemounts above? I would prefer a fixed string here, otherwise people might try to set `TLS_SECRET_NAME` and `AUTHENTICATION_CREDENTIALS_SECRET` to the same secret and we would have duplicate entry in here.

templates/operator/params.yaml.template
```diff
@@ -273,6 +273,10 @@ parameters:
     description: "Authentication backend, implementing IAuthenticator; used to identify users."
     default: "AllowAllAuthenticator"
 
+  - name: AUTHENTICATION_CREDENTIALS_SECRET
```
ANeumann82, line 276: To keep the parameter names consistent we might want to name this "AUTHENTICATION_SECRET_NAME" to be in sync with "TLS_SECRET_NAME"

operator/templates/repair-job.yaml
kensipe: agree on not using `latest`

operator/templates/node-scripts.yaml
```diff
@@ -4,10 +4,12 @@ metadata:
   name: {{ .Name }}-node-scripts
   namespace: {{ .Namespace }}
 data:
+  node-drain.sh: |
+    nodetool {{ if .Params.AUTHENTICATION_CREDENTIALS_SECRET }}-u $(cat /etc/cassandra/authentication/username) -pw $(cat /etc/cassandra/authentication/password){{ end }} drain
   node-readiness-probe.sh: |
-    nodetool status -p {{ .Params.JMX_PORT }} | grep -q "UN  ${POD_IP}"
+    nodetool {{ if .Params.AUTHENTICATION_CREDENTIALS_SECRET }}-u $(cat /etc/cassandra/authentication/username) -pw $(cat /etc/cassandra/authentication/password){{ end }} status -p {{ .Params.JMX_PORT }} | grep -q "UN  ${POD_IP}"
```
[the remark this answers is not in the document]
kensipe: agree with @ANeumann82

tests/suites/authentication/authentication_test.go
```diff
@@ -0,0 +1,111 @@
+package authentication
+
+import (
```
kensipe, line 3: this seems like it could be a kuttl test which might be easier to read and consume

operator/templates/repair-job.yaml
nfnt: I'm changed from `1.18.0` to `latest` to be consistent with us using the same image for the node resolver init container. I agree that using `latest` is not good but neither is `1.18.0` because both are assuming a specific version of the cluster. Let me fix that here and in the `Statefulset` by introducing a new parameter to set the version of this image -- the default would be `1.18`.

operator/templates/node-scripts.yaml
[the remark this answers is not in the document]
nfnt: I'll use a template variable instead of a shell variable. With this we don't have to repeat the variable definition for every command in the config map.

operator/templates/node-scripts.yaml
```diff
@@ -1,20 +1,30 @@
+{{ $auth_params := "" }}
+{{ if .Params.AUTHENTICATION_SECRET_NAME }}
```
zmalik, line 2: we should also populate the credentials for cqlsh when authentication is enabled
https://docs.datastax.com/en/security/5.1/security/secCqlshCredentials.html
otherwise `cqlsh` queries will fail

nfnt:
Passwords in env variables are fine as long as they're not exported. But you raise a good point here, because passwords as part of a command line are not fine because they can be extracted by a `ps` call. Just noticed that `TLS_KEYSTORE_PASSWORD` and `TLS_TRUSTSTORE_PASSWORD` are used like this and with cleartext password in parameters as well. I'll take a look on how to secure that. `nodetool` supports password files, let me see how this can be added.

operator/templates/node-scripts.yaml
nfnt: This should be up to the admin. We don't use `cqlsh` as a utility in the operator. The idea is that admins set up authn/authz to their liking, then provide the operator with a secret containing credentials for the operator internal use of `nodetool`.

zmalik: https://github.com/kudobuilder/operators/blob/7e4eeb595a0f8efbb73e151b26d7c0b7d9133ec9/repository/cassandra/3.11/docs/accessing.md#access-cassandra 
this won't work. So once inside pod, we should configure it accordingly so users don't have to do manual setup to gain access.

nfnt: This will work if users add `--user` and `--password` to their call :D
While I agree, that we could add the credentials of the secret to every `cqlshrc` in the cluster, this would only provide a single user for every call to `cqlsh`. Which is what authn/authz wants to avoid. Users calling `cqlsh` should have their own credentials and different authorization rules based on those credentials.

zmalik: ok yeah, you are right! 
I will mark this as resolved. I would like to see that being provided out of the box but we will provide this feature once we have any request for that and probably with another parameter 👍

nfnt:
There's no good way to fully secure `nodetool` authentication in a cluster environment. Or at least in the way we use that tool 😄. Especially the `kubectl exec -- nodetool` of the repair job is problematic. We should think of a better way to provide this functionality; a sidecar with some API endpoints could improve the situation here.

PR #91, opened 2020-04-23
Remove truststore password parameters

nfnt:
The parameter values don't need to be surfaced to users as they're only used internally.


nfnt:
More usage of the parameter in `cassandra-env-sh.yaml` and `nodetool-ssl-properties.yaml`. Will have to generate `nodetool-ssl.properties` as well.

operator/templates/generate-cassandra-yaml.yaml
```diff
@@ -235,7 +243,7 @@ data:
     # CommitLogSegments are moved to this directory on flush if cdc_enabled: true and the
     # segment contains mutations for a CDC-enabled table. This should be placed on a
     # separate spindle than the data directories. If not set, the default directory is
-    # $CASSANDRA_HOME/data/cdc_raw.
+    # \$CASSANDRA_HOME/data/cdc_raw.
```
zmalik, line 246: intended change?

operator/templates/generate-cassandra-yaml.yaml
```diff
@@ -1,10 +1,18 @@
...
-  cassandra.yaml: |
+  generate-cassandra-yaml.sh: |
+    #!/usr/bin/env bash
+
+    set -euxo pipefail
+
+    readonly truststore_password=$(cat /etc/cassandra/truststore/truststore_password)
+    readonly keystore_password=$(cat /etc/cassandra/truststore/keystore_password)
+
+    cat <<EOF >> /etc/cassandra/cassandra.yaml
```
zmalik, line 15: we want to append here or overwrite? I'm thinking about restarts caused by health check fails etc

operator/templates/generate-cassandra-yaml.yaml
nfnt: Yes, to avoid variable substitution.

operator/templates/generate-cassandra-yaml.yaml
nfnt: Overwrite. Oh, this is appending, that's a mistake. Will fix this.

templates/operator/templates/stateful-set.yaml.template
```diff
@@ -330,7 +338,9 @@ spec:
             - bash
             - -c
           args:
-            - {{ if ne $.Params.JMX_LOCAL_ONLY "true" }}
+            - /etc/cassandra/generate-cassandra-yaml.sh;
+              {{ if ne $.Params.JMX_LOCAL_ONLY "true" }}
+              /etc/cassandra/generate-nodetool-ssl-properties.sh &&
               cp /nodetool-ssl-properties/nodetool-ssl.properties /home/cassandra/.cassandra/nodetool-ssl.properties;
```
ANeumann82, line 344:
```suggestion
              cp /etc/cassandra/nodetool-ssl.properties /home/cassandra/.cassandra/nodetool-ssl.properties;
```

nfnt: Good catch! I'm was watching this very test in the CI right now. It will fail.

operator/templates/tls-store-credentials.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ $.Name}}-tls-store-credentials
+type: Opaque
+data:
+  keystore_password: Y2Fzc2FuZHJh
+  truststore_password: Y2Fzc2FuZHJh
```
ANeumann82, line 8: So, do we plan to make this some random string, or can this secret be configured by the user, or is this expected to be static?

ANeumann82: We could try to use sprigs uuid: 

{{ uuidv4 }}

nfnt: Random string would make sense. Sprig has all functions necessary for this. Though there could be implications regarding updates that we should look into first. To make it configurable by users not so much, because that's what's being removed by this PR.

nfnt: `{{ b64enc randAlphaNum 8 }}` could be used to randomize. Just needs to be Base64-encoded.

ANeumann82: Mmmmmm, but that would be regenerated every time the resource is deployed, right? That would be... not good :D

nfnt: Exactly, that's why I didn't look into this as part of this PR.

PR #92, opened 2020-04-28
Operator Cleanup

ANeumann82:
- Use only single rbac for pod exec
- Added cleanup for repair pod plan
- Added test for repeated pod repair
- Added backoffLimit: 0 to repair job
- Added test for failing repair job
- Cleaned up plan and phase names


operator/operator.yaml
```diff
@@ -85,15 +94,18 @@ plans:
...
-  repair-pod:
+  repair:
     strategy: serial
     phases:
-      - name: nodes
-        strategy: parallel
+      - name: repair
+        strategy: serial
         steps:
+          - name: cleanup
```
nfnt, line 103: 👏

operator/templates/node-exec-rbac.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
-  name: {{ .Name }}-node-backup-role
+  name: {{ .Name }}-node-exec-role
```
zmalik, line 4: can we rename it to an even more generic name? {{ .Name }}-node-role, same for sa
with bootstrap feature, we need the permissions to read/update the configmap in our namespace. Instead of adding another sa I think we can enhance one based on enabled features

ANeumann82: Yeah, i guess it makes sense. I don't think the bootstrap feature can be disabled, right? Otherwise i'd say we should use a separate rbac set. But if it's always on, we should use the same set.

I've renamed the role/rb/sa

PR #94, opened 2020-04-29
add bootstrap binary to manage IPs

zmalik:
This PR 
- add bootstrap logic to maintain the IPs list in a CM
- adds unit tests for some functions in the bootstrap code
- adds KUTTL tests as integration tests for the bootstrap logic


images/bootstrap/pkg/service/cassandra.go
```diff
@@ -0,0 +1,169 @@
...
+	// create file if not exists
+	if os.IsNotExist(err) {
+		var file, err = os.Create(replaceFile)
+		if err != nil {
+			return err
+		}
+		defer file.Close()
+	}
+	return nil
+}
```
ANeumann82, line 104: Is this func used anywhere?

images/bootstrap/pkg/service/cassandra.go
```diff
@@ -0,0 +1,169 @@
...
+	CMService *ConfigMapLock
+}
+
+func NewCassandraService(client *kubernetes.Clientset) *CassandraService {
+	return &CassandraService{
+		CMService: &ConfigMapLock{client},
+	}
+}
+
+func (c *CassandraService) SetReplaceIP() (bool, error) {
```
ANeumann82, line 34: I think this func should contain at least some sort of retry-functionality. I don't think it will happen too often, but it will happen that two nodes try to acquire/write to the config map at the same time and one of them will fail at some point.

images/bootstrap/pkg/service/cassandra.go
zmalik: not anymore it was used to generate replace file, but that was merged into WriteReplaceIp func

images/bootstrap/pkg/service/cm_lock.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	if cfg.Annotations == nil {
+		return false
+	}
+	return cfg.Annotations[annotationLock] == pod
+}
+
+func (c *ConfigMapLock) ReleaseLock() bool {
+	if c.HasLock() {
+		cfg, err := c.CoreV1().ConfigMaps(namespace).Get(configmap, meta_v1.GetOptions{})
```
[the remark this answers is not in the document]
zmalik: the lock was implemented with `re-entrant mutex` lock pattern, so we discard any option of having a deadlock here. But I understand what you mean here and it can be simplified in some cases here. Let me take a stab here!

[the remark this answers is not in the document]
zmalik: I have optimized the lock logic altogether and this as part of the improvement

images/bootstrap/pkg/service/cassandra.go
zmalik: the issue here is we probably want it to fail early. Because this is bootstrap binary which runs in background. If we don't acquire its better to do an exit(1) and restart. 
And we cannot make Cassandra process wait either as we need `nodetool status` to see if replaced IP is `UN/UJ` also. Let's sync about this on how we could probably improve this without any perf issues.

images/Dockerfile
```diff
@@ -1,3 +1,10 @@
+FROM golang:1.14.2-stretch AS bootstrap-bin
```
nfnt, line 1: Is there a reason to use this specific version? Wouldn't `golang:1.14` be enough here?

images/bootstrap/go.mod
```diff
@@ -0,0 +1,28 @@
...
+	github.com/sirupsen/logrus v1.6.0
+	github.com/spf13/pflag v1.0.5 // indirect
+	github.com/stretchr/objx v0.1.1 // indirect
+	github.com/stretchr/testify v1.3.0
+	github.com/yanniszark/go-nodetool v0.0.0-20191206125106-cd8f91fa16be // indirect
+	golang.org/x/crypto v0.0.0-20200429183012-4b2356b1ed79 // indirect
+	golang.org/x/sync v0.0.0-20200317015054-43a5402ce75a // indirect
+	golang.org/x/sys v0.0.0-20200501145240-bc7a7d42d5c3 // indirect
+	gotest.tools/gotestsum v0.4.2 // indirect
+	k8s.io/api v0.0.0-20191016110408-35e52d86657a
```
nfnt, line 25: These versions don't look right. There are tagged versions for all Kubernetes components. E.g., take a look at KUDO's `go.mod` which is using `v0.17.3` for `k8s.io/api`, `k8s.io/apimachinery` and `k8s.io/client-go`.

images/Dockerfile
zmalik: just to make sure that its a debian distro. `golang:1.14` as for today provides that but there is no guarantee for that

images/bootstrap/go.mod
zmalik: https://github.com/mesosphere/kudo-kafka-operator#go-modules you can always get the versions by kubernetes versions

images/bootstrap/main.go
```diff
@@ -0,0 +1,40 @@
...
+				os.Exit(1)
+			}
+		case "init":
+			// bootstrap to fetch the replace IP
+			if err := cassandraService.SetReplaceIPWithRetry(); err != nil {
+				log.Errorf("could not run the cassandra bootstrap: %v\n", err)
+				os.Exit(1)
+			}
+		default:
+			log.Errorf("unrecognized command for cassandra bootstrap")
```
ANeumann82, line 35:
```suggestion
			log.Errorf("unrecognized command '%s' for cassandra bootstrap", command)
```

images/bootstrap/pkg/service/cassandra.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func (c *CassandraService) SetReplaceIPWithRetry() error {
+	return retry.Do(c.SetReplaceIP, retry.Delay(RETRY_DELAY), retry.Attempts(RETRY_ATTEMPTS))
+}
+
+func (c *CassandraService) SetReplaceIP() error {
+	cfg, err := c.CMService.GetConfigMap(namespace, configmap)
+	if errors.IsNotFound(err) {
+		log.Errorf("bootstrap: configmap cassandra-topology configmap %s could not be found\n", configmap)
```
ANeumann82, line 47:
```suggestion
		log.Errorf("bootstrap: cassandra-topology configmap %s could not be found\n", configmap)
```

images/bootstrap/pkg/service/cassandra.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		log.Errorf("bootstrap: error joining the cluster with replace ip: %v\n", err)
+		return err
+	}
+
+	if err := retry.Do(c.CMService.UpdateCM, retry.Delay(RETRY_DELAY), retry.Attempts(RETRY_ATTEMPTS)); err != nil {
+		log.Errorf("bootstrap: error updating the configmap with replace ip: %v\n", err)
+		return err
+	}
+	log.Infoln("bootstrap: updating the configmap with replace ip")
```
ANeumann82, line 149:
```suggestion
	log.Infoln("bootstrap: reset replace ip")
```

images/bootstrap/pkg/service/cassandra.go
```diff
@@ -0,0 +1,159 @@
...
+		log.Errorf("bootstrap: missing BOOTSTRAP_TIMEOUT param: %v\n", err)
+		return err
+	}
+	err = c.WaitforReplacement(wait * time.Minute)
+	// re-joining can take really long time depending on the data
+	if err != nil {
+		log.Errorf("bootstrap: error joining the cluster with replace ip: %v\n", err)
+		return err
+	}
+
```
ANeumann82, line 144:
```suggestion
    log.Infoln("bootstrap: updating the configmap with new node ip")
```

images/bootstrap/pkg/service/cm_lock.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+type ConfigMapLock struct {
+	kubernetes.Interface
+}
+
+func (c *ConfigMapLock) AcquireLock() (*v1.ConfigMap, error) {
+	cfg, err := c.CoreV1().ConfigMaps(namespace).Get(configmap, meta_v1.GetOptions{})
+	if errors.IsNotFound(err) {
+		log.Warnf("bootstrap: configmap cassandra-topology configmap %s cannot be found...", configmap)
```
ANeumann82, line 20:
```suggestion
		log.Warnf("bootstrap: cassandra-topology configmap %s cannot be found...", configmap)
```

images/bootstrap/pkg/service/cassandra.go
```diff
@@ -0,0 +1,159 @@
...
+	return c.WriteReplaceIp("")
+}
+
+func init() {
+	namespace = os.Getenv("POD_NAMESPACE")
+	pod = os.Getenv("POD_NAME")
+	ipAddress = os.Getenv("POD_IP")
+	configmap = os.Getenv("CASSANDRA_IP_LOCK_CM")
+	bootstrapWait = os.Getenv("BOOTSTRAP_TIMEOUT")
+}
```
ANeumann82, line 159: Nit: Can we move this to the very top, just below the var definition?

Took me a bit to figure out that these are basically constants and not global variables ;)

images/bootstrap/pkg/service/cassandra.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	log.Infoln("bootstrap: updating the configmap with replace ip")
+	return c.WriteReplaceIp("")
+}
+
+func init() {
+	namespace = os.Getenv("POD_NAMESPACE")
+	pod = os.Getenv("POD_NAME")
+	ipAddress = os.Getenv("POD_IP")
+	configmap = os.Getenv("CASSANDRA_IP_LOCK_CM")
```
ANeumann82, line 157:
```suggestion
	configmapName = os.Getenv("CASSANDRA_IP_LOCK_CM")
```

images/bootstrap/pkg/service/cassandra.go
```diff
@@ -0,0 +1,159 @@
...
+		return err
+	}
+	log.Infoln("bootstrap: updating the configmap with replace ip")
+	return c.WriteReplaceIp("")
+}
+
+func init() {
+	namespace = os.Getenv("POD_NAMESPACE")
+	pod = os.Getenv("POD_NAME")
+	ipAddress = os.Getenv("POD_IP")
```
ANeumann82, line 156:
```suggestion
	podIpAddress = os.Getenv("POD_IP")
```

images/bootstrap/pkg/service/cassandra.go
```diff
@@ -0,0 +1,159 @@
...
+		log.Errorf("bootstrap: error updating the configmap with replace ip: %v\n", err)
+		return err
+	}
+	log.Infoln("bootstrap: updating the configmap with replace ip")
+	return c.WriteReplaceIp("")
+}
+
+func init() {
+	namespace = os.Getenv("POD_NAMESPACE")
+	pod = os.Getenv("POD_NAME")
```
ANeumann82, line 155:
```suggestion
	podName = os.Getenv("POD_NAME")
```

images/bootstrap/pkg/service/cm_lock.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, err
+	}
+	if cfg.Annotations == nil {
+		return nil, fmt.Errorf("configmap %s has no annotations", configmap)
+	}
+	if cfg.Annotations[ANNOTATION_LOCK] == pod {
+		return cfg, nil
+	}
+	return nil, fmt.Errorf("%s doesn't have the lock", configmap)
```
ANeumann82, line 47:
```suggestion
	return nil, fmt.Errorf("%s doesn't have the lock in %s", pod, configmap)
```

images/bootstrap/pkg/service/cm_lock.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (c *ConfigMapLock) HasLock() (*v1.ConfigMap, error) {
+	cfg, err := c.CoreV1().ConfigMaps(namespace).Get(configmap, meta_v1.GetOptions{})
+	if errors.IsNotFound(err) {
+		log.Warnf("bootstrap: configmap cassandra-topology configmap %s cannot be found...", configmap)
+		return nil, err
+	}
+	if cfg.Annotations == nil {
+		return nil, fmt.Errorf("configmap %s has no annotations", configmap)
+	}
```
ANeumann82, line 43: Is this an error that needs to be specifically handled?

I'd probably just maek the next line `if cfg.Annotations != nil && cfg.Annotations[ANNOTATION_LOCK] == pod`

But it may be an actual unexpected state

images/bootstrap/pkg/service/cassandra.go
zmalik: moved!

PR #96, opened 2020-04-30
Started new branch for recovery controller

ANeumann82:


docs/evicting-nodes.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Mark pod for eviction
+
+Add a new label to the node to be evicted:
+
+```bash
+kubectl label pod cassandra-node-0 kudo-cassandra/evict=true
+```
+
+This will trigger the recovery controller to unlink the PV and remove the PVC,
```
nfnt, line 39: Let's add a big warning that removing the PVC can result in deletion of the PV if the PV was configured as such. Users should be aware of this implication before evicting nodes.

images/Dockerfile.recovery-controller
```diff
@@ -0,0 +1,11 @@
+FROM golang:alpine AS build-env
```
nfnt, line 1: Let's not use `latest` or `alpine` but at least a major release of Go. E.g. `golang:1.14-alpine`.

images/Dockerfile.recovery-controller
```diff
@@ -0,0 +1,11 @@
...
+ADD ./cassandra-recovery /cassandra-recovery
+WORKDIR /cassandra-recovery
+RUN apk add --no-cache gcc musl-dev git mercurial ca-certificates
```
nfnt, line 4: Isn't that already installed as part of the [Docker image](https://github.com/docker-library/golang/blob/e7fb0efd448658b759c2958744a8c59b53dcd6b3/1.14/alpine3.11/Dockerfile)? Also, `CGO_ENABLED=0`, do we still need `gcc` in that case?

images/Dockerfile.recovery-controller
```diff
@@ -0,0 +1,11 @@
...
+ENV GO111MODULE=on
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o cassandra-recovery
+
+FROM alpine:3.7
```
nfnt, line 8: You could even use `scratch` here.

images/cassandra-recovery/pkg/controller/controller.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func NewController(client *kubernetes.Clientset) *Controller {
+	return &Controller{
+		client: client,
+	}
+}
+
+func (c *Controller) Run(ctx context.Context) {
+	namespace := os.Getenv("NAMESPACE")
```
nfnt, line 41: Let's decouple controller configuration from running the controller. How about retrieving the env vars in `main.go` and using them as arguments for `Run`?

```go
func (c *Controller) Run(ctx context.Context, instance string, namespace string)
```

images/cassandra-recovery/pkg/sts/statefulset.go
```diff
@@ -0,0 +1,224 @@
...
+	"log"
+	"os"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+var evictionLabel = os.Getenv("EVICTION_LABEL")
```
nfnt, line 14: What happens if this isn't set? It would be great if the env var would be checked in `main.go` and StatefulSet processing could be configured with a string instead of an env var.

images/cassandra-recovery/pkg/sts/statefulset.go
```diff
@@ -0,0 +1,224 @@
...
+
+func Process(client *kubernetes.Clientset, item interface{}) {
```
nfnt, line 16: Ideally, we would process `runtime.Object` here.

docs/evicting-nodes.md
ANeumann82: Good point.

images/Dockerfile.recovery-controller
ANeumann82: Good point, nope :) Actually, only git needs to be installed, i've cleaned that up.

images/cassandra-recovery/pkg/sts/statefulset.go
ANeumann82: I've reworked that. I initialize this now in the `controller.go` (simply because all other env vars are read there as well) and log a startup message.

If no eviction label is set, the controller still starts but doesn't support eviction.

images/cassandra-recovery/pkg/controller/controller.go
ANeumann82: I've moved that into a separate Options

images/Dockerfile.recovery-controller
```diff
@@ -0,0 +1,11 @@
+FROM golang:1.14-alpine AS build-env
+ADD ./recovery /recovery
+WORKDIR /recovery
+RUN apk add --no-cache git
+ENV GO111MODULE=on
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o cassandra-recovery
+
+FROM debian:buster-20200422-slim
```
nfnt, line 8: Why not `scratch`? In case we need a file system, let's use `debian:buster-slim` instead, because fixing this to a specific version also means that there won't be any security fixes to the tooling if we would rebuild this image. And rebuilding with an updated `debian:buster-slim` won't break anything, because the binary doesn't depend on the file system.

images/recovery/Makefile
```diff
@@ -0,0 +1,12 @@
+
+.PHONY: goimports
+goimports:
```
nfnt, line 3: Unnecessary, `golangci-lint` includes `goimports`.

images/bootstrap/pkg/service/cassandra.go
```diff
@@ -56,12 +56,15 @@ func (c *CassandraService) SetReplaceIP() error {
...
 		// new internal ip address
 		if isBootstrapped() {
+			log.Infof("bootstrap: Node is already  bootstrapped, no need for replace IP")
 			// bootstrapped node needs no replace ip flag
 			return nil
 		} else {
+			log.Infof("bootstrap: Node is not bootstrapped, add replace ip to startup")
 			// node not bootstrapped and has an old ip address
 			return c.WriteReplaceIp(oldIp)
 		}
```
zmalik, line 70:
```suggestion
		if !isBootstrapped() {
			log.Infof("bootstrap: Node is not bootstrapped, add replace ip to startup")
			// node not bootstrapped and has an old ip address
			return c.WriteReplaceIp(oldIp)
		}
		// bootstrapped node needs no replace ip flag
		log.Infof("bootstrap: Node is already  bootstrapped, no need for replace IP")
```

images/recovery/pkg/sts/statefulset.go
```diff
@@ -0,0 +1,230 @@
+package sts
+
+import (
+	"fmt"
+	"log"
```
zmalik, line 5: lets use logrus here also or use log in cassandra.go

images/recovery/pkg/sts/statefulset.go
```diff
@@ -0,0 +1,230 @@
...
+
+	if needsRecovery {
+		log.Printf("the pod %s/%s meets the recovery conditions.\n", pod.Namespace, pod.Name)
+		if err = cleanStartPod(client, pod); err != nil {
+			log.Printf("ERROR: Failed to clean start pod: %v", err)
+		}
+		return
+	}
+
+	if detectEvictionCondition(evictionLabel, pod) {
```
zmalik, line 40: let's move evictionCondition before `needsRecovery`. So a user added label always takes preference over recovery conditions

images/recovery/pkg/sts/statefulset.go
```diff
@@ -0,0 +1,230 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func detectRecoveryConditions(client *kubernetes.Clientset, pod *corev1.Pod) (bool, error) {
+	isUnschedulable := false
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type == "PodScheduled" && condition.Reason == "Unschedulable" {
```
zmalik, line 65:
```suggestion
		if condition.Type == corev1.PodScheduled && condition.Reason == corev1.PodReasonUnschedulable {
```

ANeumann82: Good catch, thanks!

images/Dockerfile.recovery-controller
ANeumann82: Because I failed to make it work with scratch. Figured it out now, there's a subtle but important different between:
```
CMD /cassandra-recovery/cassandra-recovery
```
and
```
CMD [ "/cassandra-recovery/cassandra-recover" ]
```

The former requires bash to be installed...

Changed it to scratch

PR #97, opened 2020-04-30
Support Cassandra clusters spanning multiple Kubernetes clusters

nfnt:
Instances can be configured to consume external seed node addresses.


nfnt:
The tests still failing due to a race when deleting namespaces, will fix this. Other than that, this PR is ready for review.

docs/multidatacenter.md
```diff
@@ -142,6 +142,18 @@ Then the cassandra node distribution would probably end up similar to this:
 - 4x `us-west-2b`
 - 4x `us-west-2c`
 
+## Adding instances running in other Kubernetes clusters
+
+Datacenters can also span multiple Kubernetes clusters. To let an instance know about a datacenter running in another Kubernetes cluster, the `EXTERNAL_SEED_NODES` parameter has to be set. This parameter takes an array of DNS names or IP addresses that seed nodes outside of the cluster have. The clusters have to be set up so that the pods running Cassandra nodes can communicate with each other.
```
ANeumann82, line 147: This needs a hint that the CLUSTER_NAME needs to be the same (I think)

nfnt: Good point! I didn't check that yet, because I was using the same cluster name in all my tests. Will add this here.

tests/suites/backup_restore/backup_restore_test.go
```diff
@@ -288,7 +288,7 @@ var _ = Describe("backup and restore", func() {
 	})
 
 	// This test is disabled (PIt instead of It) and can be enabled as soon as https://github.com/thelastpickle/cassandra-medusa/pull/119 is merged and released
-	PIt("Creates and restores a backup with JMX SSL and authentication", func() {
+	It("Creates and restores a backup with JMX SSL and authentication", func() {
```
ANeumann82, line 291: This was intentional left as `PIt`. Backup/Restore with SSL doesn't work yet, it's waiting for a new medusa release.

nfnt: Thanks, I thought this was by mistake... that's why the tests fail 😆

PR #98, opened 2020-05-04
Divorce docker.sh from hardcoded WORKDIR.

porridge:
Make the wrapper take the argument relative to $PWD rather than the repo
root.  This is more transparent and intuitive if you need to run it from
elsewhere than the repo root (like we do in some CI setups).


porridge:
FTR the `fetch-dispatchfile` failure will be fixed by https://github.com/mesosphere/kudo-cassandra-operator/pull/100

PR #99, opened 2020-05-04
Fix and enforce formatting

porridge:


porridge:
FTR the `fetch-dispatchfile` failure will be fixed by https://github.com/mesosphere/kudo-cassandra-operator/pull/100

PR #107, opened 2020-05-12
Use a random backup prefix.

porridge:
Using a hardcoded string or depending on TeamCity-specific environment
variables is not robust and hard to port.


porridge:
It's still possible to see if the data was uploaded. The prefix is printed to test output:
```
Running Suite: backup-restore-test
==================================
Random Seed: 1589283246
Will run 2 of 3 specs

Using backup prefix 17335f6f-be6f-47df-b034-913f3a12cc88
```

PR #113, opened 2020-05-15
Use locally build kuttl with 0.8.1 kind

ANeumann82:
This PR has a locally build KUTTL that uses a 0.8.1 kind. Should probably not be merged, it's just to investigate if this is the problem.


ANeumann82:
Not needed anymore, superseeded by #156

PR #114, opened 2020-05-18
More fixes for node replace

ANeumann82:
- Do not delete PVC if the Phase is still pending
- Support for JMX/SSL in bootstrap
- use `kubectl node drain` before deleting the kubernetes node


operator/templates/stateful-set.yaml
```diff
@@ -585,11 +590,27 @@ spec:
         - name: bootstrap
           image: {{ $.Params.NODE_DOCKER_IMAGE }}
           imagePullPolicy: {{ $.Params.NODE_DOCKER_IMAGE_PULL_POLICY }}
+          securityContext:
+            capabilities:
+              add:
+                # TODO(mpereira): add an operator parameter for "memory lock" so
+                # that this capability and a respective Docker image with
+                # `setcap`ed binaries are conditionally used.
+                - IPC_LOCK
```
zmalik, line 599: @ANeumann82 why we need the `IPC_LOCK` in bootstrap?

operator/templates/stateful-set.yaml
```diff
@@ -585,11 +590,27 @@ spec:
...
+                # FIXME(mpereira): might not be necessary given DCOS-59219.
+                - SYS_RESOURCE
```
zmalik, line 601: same here

kuttl-tests/Makefile
```diff
@@ -11,44 +12,55 @@ endif
 
 export PATH := $(shell pwd)/bin:$(PATH)
 export ROOT_DIR=$(realpath -L ./..)
+export DOCKER_API_VERSION=1.39
```
ANeumann82, line 15: I guess we're using the docker host from the AMI/Build Image? At least i had to force the version down to 1.39. Otherwise we'd use a client that defaults to 1.40, which fails

kuttl-tests/Makefile
```diff
@@ -11,44 +12,55 @@ endif
...
 
 .PHONY: install-kuttl
-install-kuttl: bin/kubectl-kuttl bin/kubectl-kudo bin/kubectl
+install-kuttl: bin/kubectl-kuttl bin/kubectl-kudo bin/kubectl bin/kind
 
 .PHONY: kind-test
 kind-test: install-kuttl render-test-templates
 	mkdir -p kuttl-dist/
 	go get github.com/jstemmer/go-junit-report
+	kind delete cluster
```
ANeumann82, line 63: There seem to be some cases where the kind cluster continues to exist over builds - I think we don't get a new build node each time, they are reused. To be sure, I've added the kind delete cluster here.

kuttl-tests/suites/failure-recovery/node-failure/delete-node-for-pod.sh
```diff
@@ -6,6 +6,11 @@ NAMESPACE=$3
...
 
-echo "Delete node $NODE_NAME"
+echo "Cordon node $NODE_NAME"
+kubectl cordon ${NODE_NAME}
+
+echo "Deleting pod {$POD_NAME}"
+kubectl delete pod ${POD_NAME} -n ${NAMESPACE}
 
+echo "Delete node $NODE_NAME"
 kubectl delete node ${NODE_NAME}
```
ANeumann82, line 16: This seems to be the most working version right now - Cordon, delete the pod, then delete the node.

Draining doesn't work, just deleting the node leaves the C* process running sometime...

kuttl-tests/suites/failure-recovery/node-failure/params.yaml.template
```diff
@@ -1,19 +1,20 @@
...
 NODE_MEM_MIB: "1024"
 NODE_MEM_LIMIT_MIB: "1384"
 PROMETHEUS_EXPORTER_ENABLED: "false"
 NODE_COUNT: 2
-NODE_READINESS_PROBE_INITIAL_DELAY_S: 30
-NODE_LIVENESS_PROBE_INITIAL_DELAY_S: 60
+NODE_READINESS_PROBE_INITIAL_DELAY_S: 60
+NODE_READINESS_PROBE_PERIOD_S: 10
+NODE_READINESS_PROBE_FAILURE_THRESHOLD: 8
+NODE_LIVENESS_PROBE_INITIAL_DELAY_S: 140
```
ANeumann82, line 13: The replace bootstrap process can take a bit - there is an initial ring join delay that's 30 seconds by default, so we want to give it enough time to actually replace the node

kuttl-tests/suites/kuttl-failure-recovery.yaml.template
```diff
@@ -13,5 +13,5 @@ kindContainers:
   - {{ .CASSANDRA_DOCKER_IMAGE }}
   - {{ .RECOVERY_CONTROLLER_DOCKER_IMAGE }}
 {{- end }}
-timeout: 300
+timeout: 360
```
ANeumann82, line 16: 300 seconds might be enough, but above timeouts are around 140 seconds per node, leaving only 20 seconds to ensure everything is up and running. 360 seconds seems to be a bit safer.

operator/templates/cassandra-env-sh.yaml
```diff
@@ -251,8 +251,9 @@ data:
 
     {{ if ne $.Params.JMX_LOCAL_ONLY "true" }}
       JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.local.only=false"
-      JVM_OPTS="$JVM_OPTS -Djava.rmi.server.hostname=$POD_NAME.{{ .Name }}-svc.$POD_NAMESPACE.svc.cluster.local"
-      
+      #JVM_OPTS="$JVM_OPTS -Djava.rmi.server.hostname=$POD_NAME.{{ .Name }}-svc.$POD_NAMESPACE.svc.cluster.local"
+      JVM_OPTS="$JVM_OPTS -Djava.rmi.server.hostname=$POD_IP"
+
```
ANeumann82, line 256: This might get undone later, it's to try another version to shutdown the old cassandra instance

operator/templates/generate-cassandra-yaml.yaml
```diff
@@ -46,17 +46,20 @@ data:
...
 
     # When replacing a node, the new node should not be a seed node either. This applies to the very first node as well
     if [ -s /var/lib/cassandra/replace.ip ]; then
         echo "File replace.ip found, we are in replacement mode and have to remove our own host from seed list";
         SEEDS=`echo $SEEDS | sed -E 's/'"$HOSTNAME"',?//'`
     fi;
+    set +x
+
+    # Remove trailing comma if there is one
+    SEEDS=`echo $SEEDS | sed -E 's/,?$//'`
```
ANeumann82, line 62: Trailing comma "works", but spams the Cassandra log because C* interprets the last empty string as "localhost" and tries to add itself, which fails.... Pretty annoying

operator/templates/stateful-set.yaml
ANeumann82: https://github.com/mesosphere/kudo-cassandra-operator/blob/3884e69e01c19ed80e01b0c69ee76572fb245e44/images/Dockerfile#L23

Because of this, basically. We give the java binary the capabilities, so when we try to run java (i.e. nodetool), and the container doesn't have the caps, it fails.

operator/templates/stateful-set.yaml
```diff
@@ -585,11 +582,32 @@ spec:
...
             - -c
           args:
-            - /etc/cassandra-bootstrap/bootstrap init
+            - {{ if or (eq $.Params.TRANSPORT_ENCRYPTION_ENABLED "true") (eq $.Params.TRANSPORT_ENCRYPTION_CLIENT_ENABLED "true") (ne $.Params.JMX_LOCAL_ONLY "true") }}
+              /etc/tls/bin/generate-tls-artifacts.sh;
+              {{ end }}
+              {{ if ne $.Params.JMX_LOCAL_ONLY "true" }}
+              /etc/cassandra/generate-nodetool-ssl-properties.sh &&
+              cp /etc/cassandra/nodetool-ssl.properties /home/cassandra/.cassandra/nodetool-ssl.properties;
+              {{ end }}
```
ANeumann82, line 604: We need to have nodetool-ssl-properties here already, same for the tls artifacts

operator/templates/stateful-set.yaml
```diff
@@ -585,11 +582,32 @@ spec:
...
+              {{ if $.Params.NODE_TOPOLOGY }}
+              /etc/cassandra/wait-for-node-zero.sh {{ $.Name }}-{{ $datacenter.datacenter }};
+              {{ else }}
+              /etc/cassandra/wait-for-node-zero.sh {{ $.Name }};
+              {{ end }}
```
ANeumann82, line 609: Much better to wait here for node-zero, because we don't run into readyness checks

templates/operator/templates/stateful-set.yaml.template
```diff
@@ -599,11 +599,7 @@ spec:
               cp /etc/cassandra/nodetool-ssl.properties /home/cassandra/.cassandra/nodetool-ssl.properties;
               {{ end }}
               /etc/cassandra/generate-cqlshrc.sh;
-              {{ if $.Params.NODE_TOPOLOGY }}
```
zmalik: 👏 thanks for simplifying that part

PR #116, opened 2020-05-20
bump cassandra to 3.11.6

zmalik:
use the latest Cassandra 3.11.6


zmalik:
@nfnt for the stress tool image we would need a release, as its just used to stress against any Cassandra we can bump that later

PR #123, opened 2020-05-26
Add nodeAffinity label to cordon only cassandra from a k8s node, kudo v0.13.0

ANeumann82:
Added an nodeAffinity label for nodes not to contain a specific label. This makes it a lot easier to evict nodes from a k8s node without draining or cordon the whole node.

Updated KUDO dependency to 0.13.0


metadata.sh
```diff
@@ -38,7 +38,7 @@ export VENDOR_DIRECTORY="${_project_directory}/shared/vendor"
 export CASSANDRA_VERSION="3.11.5"
 
 # https://github.com/kudobuilder/kudo/releases/tag/vx.y.z
-export KUDO_VERSION="0.11.1"
+export KUDO_VERSION="0.13.0"
```
zmalik, line 41: @ANeumann82 can you also please bump in here https://github.com/mesosphere/kudo-cassandra-operator/blob/master/kuttl-tests/Makefile#L1 ?

PR #124, opened 2020-05-27
Update Medusa to 0.6.0

ANeumann82:
This allows us to enable backup/restore for encrypted clusters


tools/docker.sh
```diff
@@ -13,4 +13,6 @@ fi
 
 docker build "${REPO_ROOT}/tools" -t kudo-cassandra-tools
 
-docker run --rm -u "$(id -u):$(id -g)" -v "${REPO_ROOT}:${REPO_ROOT}" -w "$(pwd)" kudo-cassandra-tools "$@"
+
+# Run docker and copy all ENV vars except any that contain PATH
+docker run --rm -u "$(id -u):$(id -g)" --env-file <(env | grep -v "PATH") -v "${REPO_ROOT}:${REPO_ROOT}" -w "$(pwd)" kudo-cassandra-tools "$@"
```
porridge, line 18: I'm a little concerned about this particular change, can you explain why you needed it?
First, I think the list of variables is too broad and I'm afraid it will cause  problems in the future. I'd rather see variables passed explicitly when needed by whatever invokes this script, with something like `./tools/docker.sh env "VAR1=$VAR1" "VAR2=$VAR2" ./something/or/other.sh`
Second, the `grep` looks at values, but probably should not.

ANeumann82: Well... It's a problem with docker-ception and running every nook and cranny inside docker. The problem is that we need a lot of stuff from `metadata.sh`, which in turn needs a lot of other variables.
I agree that this is not optimal, but tbh. i would prefer to not have the `docker.sh` at all in the cassandra project. I think it should rather be in the shared tools, pass all required vars from TC to that docker invocation and let the project do it's stuff inside there.

ANeumann82: Also: In this case the `docker.sh` is called from the `image-build-and-push.sh` which is located in the shared directory...

I have adjusted the pattern to only look for PATH in the var name

porridge:
#134 and #135 are merged, FWIW.

tools/docker.sh
ANeumann82: I've removed this after merging master

run-tests.sh
```diff
@@ -11,7 +11,9 @@ source "${project_directory}/metadata.sh"
 
 if [[ -n ${IMAGE_DISAMBIGUATION_SUFFIX:-} ]]; then
   # Refresh templated files to pick up the suffix, if explicitly set.
-  "${project_directory}/tools/docker.sh" "${project_directory}/tools/compile_templates.sh"
+  "${project_directory}/tools/docker.sh" \
```
porridge, line 14: :man_facepalming: 
Found like you said we would :laughing:

PR #130, opened 2020-06-02
Dispatch ccc

faiq:


Dispatchfile
```diff
@@ -27,21 +29,16 @@ task("goimports-check", inputs=["kudo-cassandra-operator"], steps=[k8s.corev1.Co
   )
 )])
 
-
```
orsenthil, line 30: Not your change> but I observe that `goimports` step is requesting for
> 7.5Gi
of memory.

PR #131, opened 2020-06-02
Disable the Prometheus exporter by default

nfnt:
The Prometheus exporter requires additional CRDs that aren't available in many default cluster environments. To avoid errors in this case, the `PROMETHEUS_EXPORTER_ENABLED` has been changed from opt-out to opt-in.


docs/parameters.md
```diff
@@ -1,246 +1,245 @@
...
-| **REPAIR_POD**                                           | Name of the pod on which 'nodetool repair' should be run.                                                                                                                                                                                                                                                                                 |                                                                                                                                                                                                   |
+| [1m                          Name                          [0m |                                                                                                                                                                Description                                                                                                                                                                |                                                                                              Default                                                                                              |
```
nfnt, line 2: Looks like some additional formatting was added to the table by an updated `pytablewriter`. Not sure if this is what we want. My environment is using version 0.54.0.

docs/monitoring.md
```diff
@@ -8,8 +8,7 @@ The KUDO Cassandra operator will export metrics to Prometheus by default. It
 achieves this using a Prometheus exporter based on the
 [criteo/cassandra_exporter](https://github.com/criteo/cassandra_exporter).
 
-When the `PROMETHEUS_EXPORTER_ENABLED` parameter is at its default value of
-`true`:
+When the `PROMETHEUS_EXPORTER_ENABLED` parameter is set to `true`:
```
porridge, line 11: Please correct line 7 as well as 44/45.

docs/parameters.md
porridge: I recall @zmalik said something about changing the version back a little, not sure it happened.

nfnt: The CI uses 0.51.0, I downgraded locally.

porridge:
@nfnt please read the entirety of section https://github.com/mesosphere/kudo-cassandra-operator/pull/131/files#diff-609282fcf45c95a88e8b87096f80263aR44-R58 and re-write it to make sense. This guide was written for people who want this enabled so sentence such as "you only need to worry about this if you enabled it" has little sense :-)

There is at least one other sentence in this paragraph which assumes the previous default. Please fix appropriately.

zmalik:
@nfnt we are testing metrics in the sanity test can you enable monitoring for sanity test?

nfnt:
Thanks for catching and correcting that @zmalik!

docs/monitoring.md
```diff
@@ -42,8 +41,8 @@ namespace_name=default
 
 ### 1. Make sure that Prometheus Exporter is enabled on the KUDO Cassandra instance
 
-This parameter is `true` by default, so you only need to worry about this if you
-explicitly disabled it.
+This parameter is `false` by default, so you only need to worry about this if
+you already have a KUDO Cassandra instance running.
```
porridge, line 45:
```suggestion
This parameter is `false` by default, so you need to enable it explicitly.
```

docs/monitoring.md
```diff
@@ -42,8 +41,8 @@ namespace_name=default
...
 
 If you do not remember, you can check the value of the parameter on a running
```
porridge, line 47:
```suggestion
You can check the value of the parameter on a running
```

PR #138, opened 2020-06-02
Switching to kuttl for MWT

kensipe:
Switch to using KUTTL for MWT

I've tested against a Konvoy cluster with the following changes to the `setup/01-cassandra-install/02-install-cassandra.yaml` file

```
  parameters:
    NODE_COUNT: "3"
    NODE_MEM_MIB: "4096"
    NODE_MEM_LIMIT_MIB: "4096"
    NODE_CPU_MC: "1000"
    NODE_CPU_LIMIT_MC: "1000"
    NODE_DISK_SIZE_GIB: "20"
```

I would be inclined to restructure the folders and README if we can get agreement on this is the way we want to move forward.  (I would likely remove or rename `mwt` folder and merge the 2 README files)

I did discover some weakness in kuttl with this... namely around namespace management... the `delete` properties for a TestStep do not honor the namespace defined... they only delete what is in the generated test namespace.  Which has worked well for tests but isn't great when the focus is setup and verify.  Created https://github.com/kudobuilder/kuttl/issues/124 to track.

Signed-off-by: Ken Sipe <kensipe@gmail.com>


benchmarks/mwt/setup/00-kudo-check/00-assert.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  labels:
+    app: kudo-manager
+  name: kudo-controller-manager-0
+  namespace: kudo-system
+status:
+  containerStatuses:
+  - name: manager
+    ready: true
+    image: docker.io/kudobuilder/controller:v0.14.0-rc1
```
zmalik, line 12: can we template versions somehow? 🤔 this always changes between MWTs and same goes for `operatorversion` `cassandra-0.1.2`

benchmarks/mwt/setup/01-cassandra-install/01-install-crd.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - command: kubectl kudo install cassandra --skip-instance --namespace cassandra
```
zmalik, line 4: we should test from the local path in this case if MWT scripts are living next to operator

benchmarks/mwt/setup/00-kudo-check/00-assert.yaml
kensipe: I debated this... we can just not include the container image and we will be good.   It depends on how exacting we want to be ...   I'll remove it for now

benchmarks/mwt/setup/01-cassandra-install/01-install-crd.yaml
kensipe: @zmalik I don't know what you mean

kensipe: if you are referring to using the local build of this operator in this repo... I would rather not... MWT should only be on released operators and the scripts should work against the released operator repo IMO

ANeumann82: I would generally agree, although with a project that is heavily in development this might not always be possible. For the last MWT I used a dev-version from master.

I do hope that we have reached a state where this won't be necessary anymore though.

zmalik: we usually participate with the next unreleased version in MWT to make sure there are no stability issues with the upcoming release under load tests.
that's a non-blocker issue as would require minimal effort to change it when running MWT. Moreover, now given the new dates we will participate with a released version so each time it will heavily dependent on operator release dates and MWT dates

benchmarks/mwt/README.md
```diff
@@ -0,0 +1,48 @@
...
+* Kubernetes cluster up
+* KUDO CLI installed on node kuttl is running from
+* KUDO manager installed in the cluster
+
+
+# TestSuites
+
+* **Setup:** used to setup a MWT test use the following command 
+
+`kubectl kuttl test setup/  --parallel 1 --skip-delete`
```
ANeumann82, line 24: I'd like to see a kuttl TestSuite file that has these parameters. Having them as parameters and to remember to add them seems like an easy cause of failure

benchmarks/mwt/setup/01-cassandra-install/02-install-cassandra.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: kudo.dev/v1beta1
```
ANeumann82, line 1: Any specific reason why we're not using the kudo cli here? In the other kuttl tests I've usually used something like

```
apiVersion: kudo.dev/v1beta1
kind: TestStep
commands:
  - command: kubectl kudo install --instance cassandra cassandra  -P params.yaml
    namespaced: true
```

and an extra `params.yaml` for the actual parameters. Makes it a lot easier for structured parameters as "NODE_TOPOLOGY" which is a yaml structure.

benchmarks/mwt/README.md
kensipe: that is easily done... however that would require a kuttl-test.yaml for 2+ sets of tests... the setup and the teardown... which means need to run `kubectl kuttl test --config kuttl-setup.yaml` and  `kubectl kuttl test --config kuttl-teardown.yaml`

So there are still things to "know"... but it does push out config details... (which I like a lot).  I'll add this in.

benchmarks/mwt/setup/01-cassandra-install/02-install-cassandra.yaml
kensipe: no reason... I prefer to work within kuttl as much as possible and not shelling out with commands... but both ways are supported.

kensipe:
after fixing kudo... might want to look at
for 02-install-cassandra.yaml
```
apiVersion: kuttl.dev/v1beta1
kind: TestStep
commands:
  - command: kubectl kudo install --instance cassandra cassandra --namespace cassandrra  -P params.yaml    
```

kensipe:
updated to cassandra 1.0.0 + added test configs

PR #141, opened 2020-06-03
Add v1 dashboard

zmalik:
This PR:

- adds a v1 dashboard and removes SLA and adds `HEALTH` section 
- all dashboard uses `namespace` in their queries so we don't mix different namespace metrics. Move `cluster` above `datacenter` variable 
- adds a simple workload.yaml file to test with `kubectl kudo install cassandra` in `default` namespace 

Dashboard:

image


zmalik:
thanks @porridge I think this is a great idea and I will update the dashboard

PR #150, opened 2020-06-08
Add polling to ext service test

### LINKED ISSUE #146: external_service_test.go does not wait for service to disappear
The test for disabling the service expects the service to be gone right after the deploy plan finishes.

This is incorrect. In many cases the service resource is deleted but still present due to a finalizer. Example:

```
- apiVersion: v1
  kind: Service
  metadata:
    annotations:
      kudo.dev/last-applied-configuration: |
   [...]
      kudo.dev/last-plan-execution-uid: 9285b6a6-55ec-49a6-a6cb-02c867646b45
      kudo.dev/phase: nodes
      kudo.dev/plan: deploy
      kudo.dev/step: pre-node
    creationTimestamp: "2020-06-04T17:00:10Z"
    deletionGracePeriodSeconds: 0
    deletionTimestamp: "2020-06-04T17:01:38Z"
    finalizers:
    - service.kubernetes.io/load-balancer-cleanup
    labels:
      heritage: kudo
      kudo.dev/instance: cassandra-instance
      kudo.dev/operator: cassandra
    name: cassandra-instance-svc-external
    namespace: ext-service-test-namespace
    ownerReferences:
    - apiVersion: kudo.dev/v1beta1
      blockOwnerDeletion: true
      controller: true
      kind: Instance
      name: cassandra-instance
      uid: 7751f32b-67bb-428f-a5ad-0fe0317509f1
    resourceVersion: "19404"
    selfLink: /api/v1/namespaces/ext-service-test-namespace/services/cassandra-instance-svc-external
    uid: cfe3ec6f-754d-4772-b877-e1485f8e862c
```

ANeumann82:
Fixes #146 

### DIFF
--- a/tests/suites/external-service/external_service_test.go
+++ b/tests/suites/external-service/external_service_test.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"strconv"
 	"testing"
+	"time"
 
 	"github.com/kudobuilder/test-tools/pkg/client"
 	"github.com/kudobuilder/test-tools/pkg/debug"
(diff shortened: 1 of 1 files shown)


ANeumann82:
> My only concern is that "false is not true" is not a great message when a test fails.
> The `Infof`s do help, but they have the downside of spamming the output.
> I wonder whether we could come up with something even better? Perhaps return a structured object from the `func` and do a more complex assertion (which hopefully will have a more meaningful error message)?

I mostly agree. I tried to figure something out, but it's all pretty annoying. The structured object return might be an option, but it adds a lot of complexity. 
I don't think the log spam is too much, i've added a bit more detailed logs and removed the spammy output that was there before.

PR #151, opened 2020-06-09
Create simple Operator for Workload generation

ANeumann82:
Added a small operator that deploys the cassandra-stress tool.
The KUTTL test deploys that operator, waits for it it finish running the workload and downloads the logs.


benchmarks/cassandra-stress/params.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - name: TARGET_HOST
+    description: "A cassandra node that is used by cassandra stress"
+    required: true
+
+  - name: PARALLEL
+    description: "Number of Cassandra Stress instances to start"
+    default: "3"
+
+  - name: THREADS_PER_CLIENT
+    description: "Number of threads per client"
```
nfnt, line 12: Nit: What's a client in this context? Is there a client per instance? Should this be `THREADS_PER_INSTANCE`?

benchmarks/cassandra-stress/templates/stress.yaml
```diff
@@ -0,0 +1,62 @@
...
+        cassandra-stress: {{ $.OperatorName }}
+    spec:
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 999
+        runAsGroup: 999
+        fsGroup: 999
+      containers:
+        - name: cassandra-stress
+          image: mesosphere/cassandra:3.11.6-1.0.0
```
nfnt, line 27: We need to ensure that this version gets updated when the version defined in `metadata.sh` is changed. Should this get templatized like it is for the Cassandra operator?

ANeumann82: Yeah... It probably should. I shy away from templating it though, it feels like an additional step somehow, but I guess it's probably for the best. Still not sure if we should add templating to kuttl, but i guess it needs to be done at some point...

benchmarks/cassandra-stress/params.yaml
ANeumann82: I've adjusted the `PARALLEL` parameter to say `NUM_CLIENTS` and changed the docs, should be more understandable now

PR #155, opened 2020-06-22
Rework Readiness and Liveness Probes

ANeumann82:


operator/templates/node-scripts.yaml
```diff
@@ -15,17 +15,13 @@ data:
...
+    IS_NATIVE_TRANSPORT_RUNNING=`curl -s localhost:{{ $.Params.JOLOKIA_PORT }}/jolokia/read/org.apache.cassandra.db:type=StorageService/NativeTransportRunning | jq .value`
+    echo "Native Transport Running: $IS_NATIVE_TRANSPORT_RUNNING"
+    [[ $IS_NATIVE_TRANSPORT_RUNNING == "true" ]]
   node-liveness-probe.sh: |
-    {{ if ne $.Params.JMX_LOCAL_ONLY "true" }}
-    nodetool {{ $auth_params }}  -p {{ .Params.JMX_PORT }} --ssl info
-    {{ else }}
-    nodetool {{ $auth_params }} -p {{ .Params.JMX_PORT }}  info
-    {{ end }}
+    DEADLOCKED_THREADS=`curl -s localhost:{{ $.Params.JOLOKIA_PORT }}/jolokia/exec/java.lang:type=Threading/findDeadlockedThreads | jq .value`
```
nfnt, line 22: I'm curious why this particular endpoint is used. Wouldn't it be sufficient to test that there is _any_ response at `localhost:$JOLOKIA_PORT`?

ANeumann82: tbh, I'm not fully sure if we need a liveness check at all - from my understanding we want to make sure that C* is still running and working correctly, otherwise we want to restart it.

There may be some JMX beans that could be used to determine that, but I don't know which ones would for certain indicate a stalled C* instance.

Having deadlocked threads is a hard indicator that the process should be restarted. Maybe we can find and add other indicators later on...

zmalik: @ANeumann82 we could use `LiveNodes` and check if current node is on the list. That would be Cassandra definition of being `live`. Its in `StorageService`  also

https://github.com/apache/cassandra/blob/dd76858c7652541c7b137323f7b9e154686d6fba/src/java/org/apache/cassandra/service/StorageServiceMBean.java#L41 

which is basically used to show in nodetool status also the`U` nodes https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/tools/nodetool/Status.java#L153
WDYT?

PR #156, opened 2020-06-30
Use Kuttl 0.5.0 (and kind 0.8.1)

ANeumann82:


kuttl-tests/Makefile
```diff
@@ -1,6 +1,6 @@
 KUDO_VERSION=0.13.0
-KUTTL_VERSION=0.2.1
-KUBERNETES_VERSION=1.17.2
+KUTTL_VERSION=0.5.0
+KUBERNETES_VERSION=1.18.2
```
nfnt, line 3: We could also use 1.18.4, it's the latest point release that also has `kindest/node` Docker images. WDYT?

PR #159, opened 2020-08-24
Fix KUDO version usage in tests

### LINKED ISSUE #160: Setting DS_KUDO_VERSION externally does not work in tests
Both, `run-pre-checks.sh`, as well as `run-tests.sh` are run as part of our TeamCity CI. There is an option to run custom builds using a specific version of KUDO. This version is injected through the `DS_KUDO_VERSION` env var. Both scripts don't support this variable and will always use a specific version of KUDO (currently 0.15-rc1).

nfnt:
The DS_KUDO_VERSION parameter wasn't honored by 'run-pre-checks.sh' and 'run-tests.sh'. This has been fixed. Most of the functionality of 'kuttl-tests/Makefile' has been extracted into a shell script that uses 'metadata.sh'.
Also removes some redundancies in the scripts.

Fixes #160

### DIFF
--- a/kuttl-tests/kind-test.sh
+++ b/kuttl-tests/kind-test.sh
@@ -15,10 +15,10 @@ KIND_VERSION=0.8.1
 
 ARTIFACTS=kuttl-dist
 
-OS=$(shell uname -s | tr '[:upper:]' '[:lower:]')
-KUDO_MACHINE=$(shell uname -m)
-MACHINE=$(shell uname -m)
-ifeq "$(MACHINE)" "x86_64"
+OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+KUDO_MACHINE=$(uname -m)
+MACHINE=$(uname -m)
+ifeq "${MACHINE}" "x86_64"
   MACHINE=amd64
 endif
 
--- a/kuttl-tests/kind-test.sh
+++ b/kuttl-tests/kind-test.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+readonly script_directory="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+readonly project_directory="$(readlink -f "${script_directory}/..")"
+
+# shellcheck source=../metadata.sh
+source "${project_directory}/metadata.sh"
+
+KUBECTL_VERSION=1.18.4
+KUTTL_VERSION=0.5.0
+KUBECTL_KUDO_VERSION=${DS_KUDO_VERSION#v}
+KIND_VERSION=0.8.1
+
+ARTIFACTS=kuttl-dist
+
+OS=$(shell uname -s | tr '[:upper:]' '[:lower:]')
+KUDO_MACHINE=$(shell uname -m)
+MACHINE=$(shell uname -m)
+ifeq "$(MACHINE)" "x86_64"
+  MACHINE=amd64
+endif
+
+mkdir -p bin/
+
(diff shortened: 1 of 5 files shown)


tests/run.sh
```diff
@@ -10,11 +10,8 @@ source "${project_directory}/metadata.sh"
 
 # Note: the following environment variables are required by the shared
 # data-services-kudo tooling.
-export DS_KUDO_VERSION="${DS_KUDO_VERSION:-v${KUDO_VERSION}}"
```
ANeumann82, line 13: Why did you remove this one? When this is gone, it basically prevents executing `run.sh` locally, because DS_KUDO_VERSION is undefined.

nfnt: DS_KUDO_VERSION is already set in `metadata.sh`. Setting it here again is redundant.

ANeumann82: Ah 👍

kuttl-tests/kind-test.sh
```diff
@@ -0,0 +1,48 @@
...
+OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+KUDO_MACHINE=$(uname -m)
+MACHINE=$(uname -m)
+if [ "${MACHINE}" == "x86_64" ]; then
+  MACHINE=amd64
+fi
+
+mkdir -p bin/
+
+curl -Lo "bin/kubectl_${KUBECTL_VERSION}_${OS}" "https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VERSION}/bin/${OS}/${MACHINE}/kubectl"
```
zmalik, line 27: this will download on each run. Why not keep them in Makefile targets? and maybe fix them if they were being downloaded on each run

nfnt: Makefiles aren't the right tool for the job here. This is more about batching, less about complex dependency management. Furthermore, Makefiles aren't great at handling environment variables and other script related things -- even though it's not impossible. And we already have a defined way of setting up the environment for scripts with `metadata.sh` and should keep using this here.
Good point about the downloads though, let me add some checks for existing binaries.

kuttl-tests/kind-test.sh
```diff
@@ -0,0 +1,48 @@
...
+ARTIFACTS=kuttl-dist
+
+OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+KUDO_MACHINE=$(uname -m)
+MACHINE=$(uname -m)
+if [ "${MACHINE}" == "x86_64" ]; then
+  MACHINE=amd64
+fi
+
+mkdir -p bin/
```
porridge, line 25: Please make the script not silently depend on correct $PWD. Prepend, or worst case - `cd` into `$script_directory`.

kuttl-tests/kind-test.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
```
porridge, line 3:
```suggestion
set -euo pipefail
```
Do we need the versbosity?

porridge (approved):

kuttl-tests/kind-test.sh
nfnt: I agree, but on the other hand that's true for many of the scripts in this project. A correct $PWD is expected. I'd love to clean this up not only here but in the other scripts as well. As part of a separate PR. WDYT?

kuttl-tests/kind-test.sh
nfnt: For consistency with the existing scripts.

kuttl-tests/kind-test.sh
porridge: Whatever works for you. Personally I'd improve the script while writing it. Too often I'd find that such future cleanup never actually happens, because there are more important things to do :-)

ANeumann82 (approved):

PR #164, opened 2020-10-30
Introduce a parameter to set annotations for the external service

nfnt:
ExternalDNS access is enabled by adding `external-dns.alpha.kubernetes.io/*` annotations to services. This can be done using the `EXTERNAL_SERVICE_ANNOTATIONS` parameter.

### DIFF
--- a/templates/operator/params.yaml.template
+++ b/templates/operator/params.yaml.template
@@ -338,6 +338,14 @@ parameters:
     default: "9160"
     group: external
 
+  - name: EXTERNAL_SERVICE_ANNOTATIONS
+    displayName: "External service annotations"
+    hint: "Annotations added to the external service."
+    type: map
+    description: "Custom annotations for the external service."
+    default:
+    group: external
+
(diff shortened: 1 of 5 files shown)


meichstedt:
Can you add docs similar to [these](https://github.com/mesosphere/kudo-kafka-operator/pull/64/files#diff-a05d074958242ef5471564479c6fbb2a83ef2f98a82998b1765d7f618de2701e) so we have some examples of what the config looks like?

templates/operator/params.yaml.template
```diff
@@ -338,6 +338,14 @@ parameters:
     default: "9160"
     group: external
 
+  - name: EXTERNAL_SERVICE_ANNOTATIONS
+    displayName: "External service annotations"
+    hint: "Annotations added to the external service."
```
ANeumann82, line 343:
```suggestion
    hint: "Key-Value Pairs of annotations."
```

ANeumann82 (approved):
looks good, the hint could be improved though.

PR #170, opened 2021-01-15
Add upgrade test in pre checks

ANeumann82:


kuttl-tests/suites/upgrade/upgrade-test/00-install.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kudo.dev/v1beta1
+kind: TestStep
+commands:
+  - command: cat old-params.yaml
+  - command: kubectl kudo install -n cassandra-upgrade-test --instance cassandra -P old-params.yaml cassandra --app-version 3.11.5 --operator-version 0.1.2
```
rishabh96b, line 5: Instead of hardcoding, we can use an env variable for Cassandra and its operator version.

rishabh96b (approved):

